PR #5, opened 2016-05-22
Attachment support

jasonob:
Added support to upload & download attachments, and an additional method to see if a session has been authenticated

I did consider whether to add a new Attachment type to Client (as a peer of IssueService) but decided against it.

### DIFF
--- a/authentication.go
+++ b/authentication.go
@@ -1,8 +1,6 @@
 package jira
 
-import (
-	"fmt"
-)
+import "fmt"
 
 // AuthenticationService handles authentication for the JIRA instance / API.
 //
--- a/authentication.go
+++ b/authentication.go
@@ -63,6 +63,15 @@ func (s *AuthenticationService) AcquireSessionCookie(username, password string) 
 	return true, nil
 }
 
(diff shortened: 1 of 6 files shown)


andygrunwald:
Thank you @jasonob 
I will try to have a look during the week, at least at the weekend, ok? I am heavily busy right now, but just wanted to let you know that i appreciate your contribution.

jasonob:
@andygrunwald - I've added some additional test cases, and fixed one error in PostAttachment() which wasn't handling the response correctly. Its proving tricky to get the test coverage up further.

andygrunwald:
Thank you Jason!
I highly appreciate your work and it was a pleasure to merge it.
I was not able to merge it directly via UI, due to some git conflicts with other PRs in the meantime.

I removed the `func DoNoClose`, because in the meantime https://github.com/andygrunwald/go-jira/pull/6 was merged. There we moved the `defer close()` only if v is applied.
In your API function you didn`t applied `v interface{}` so the standard `Do` should work here.
Do you agree me?

Thank you!

jasonob:
Andy,

The removal of DoNoClose() looks fine. 

Cheers!

Jason

PR #6, opened 2016-05-24
Minor bug fig on parsing HTTP body

KentBrockman:
Hey Andy

Noticed a small bug here - if I haven't passed a struct into this line:
[jira.go:94](https://github.com/andygrunwald/go-jira/blob/master/jira.go#L94): `func (c *Client) Do(req *http.Request, v interface{}) (*http.Response, error) {`

There wasn't a way for me to get the resulting object out, because this line:
[jira.go:100](https://github.com/andygrunwald/go-jira/blob/master/jira.go#L100): `defer resp.Body.Close()`
already closed the body even if no struct is passed.  I made this change to let me parse responses straight from the HTTP response.  I included the test.  If you want to see the red condition on the test, just revert my changes to jira.go.

I'm fairly new to pull requests and the like - let me know if you would like this pull request to come from another branch.  Or let me know if I've possibly introduced bugs in your code with this change - hopefully not!

Kent

### DIFF
--- a/jira.go
+++ b/jira.go
@@ -97,8 +97,6 @@ func (c *Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 		return nil, err
 	}
 
-	defer resp.Body.Close()
-
 	err = CheckResponse(resp)
 	if err != nil {
 		// Even though there was an error, we still return the response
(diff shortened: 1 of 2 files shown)


KentBrockman:
Oops - my bad.  Will figure out what is breaking :smiley:

KentBrockman:
Just flubbed a test case.  Let me know if that looks good or if there should be any changes.  

Kent

andygrunwald:
Thank you Kent!

PR #9, opened 2016-05-25
Issue 7 labels in issue fields

avadhutp:
1. Include labels into IssueFields
2. Tested on JIRA `v6.4.10#64025-sha1:5b8b740`
3. Markup from [here](https://issues.apache.org/jira/rest/api/2/issue/MESOS-3325) works against this


avadhutp:
Also tested against [markup](https://jira.atlassian.com/rest/api/2/issue/USER-4) from `v7.1.2#71006-sha1:8e7e309`. And that works fine too.

andygrunwald:
Thank you, Avadhut! Good work!

avadhutp:
Thanks for getting this in so quickly @andygrunwald

PR #15, opened 2016-06-07
Expose issue link direction, comment ID

joushou:
A small two-in-one PR:
* The ID field in the Comment struct was missing.
* It was hard to detect the issue link direction without checking for the presence of OutwardIssue and InwardIssue. However, as they were not pointer types, detecting their presence was not possible, and one would have to check for zero values in the resulting Issue struct, which wasn't very nice. The same goes for presence of the issue link comment.


andygrunwald:
Thank you a lot!

PR #16, opened 2016-06-14
When creating issue links, the id and self should be omitted.

tritonrc:
Also omit the comment if none is provided


andygrunwald:
Thank you!

PR #24, opened 2016-07-07
Fix Comments unmarshalling

### LINKED ISSUE #23: Comments are not unmarshalled
Using go-jira for a CLI, I found that comments are not being unmarshalled for any Issue.
I did the following try on your code, and got `issue_test` doesn't pass. In https://github.com/andygrunwald/go-jira/blob/master/issue_test.go#L140, I added:
```go
if len(issue.Fields.Comments) != 1 {
  t.Error("Expected one comment, %v found", len(issue.Fields.Comments))
}
```
I'm not very experienced in Go programming and even most not sure of understanding JSON Unmarshalling anyway, so I might be wrong. I might like to contribute a fix for this, but I'm uncertain about this actually being a bug, or how to diagnose and fix... I also found another issue, trying to see the response body, but this should probably be another issue.

I apologize if this is not the correct way to contribute, or help; I'm just trying to use this :)

ifosch:
This fixes #23 
I know this is not the most desired way of accessing Comments, since it requires to duplicate reference to Comments. I'm looking in a better way of doing this, but being a slice, it doesn't work.
I might like to find a better way of doing this, so help here would be appreciated.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -78,7 +78,7 @@ type IssueFields struct {
 	AggregateProgress *Progress     `json:"aggregateprogress,omitempty"`
 	Worklog           *Worklog      `json:"worklog,omitempty"`
 	IssueLinks        []*IssueLink  `json:"issuelinks,omitempty"`
-	Comments          []*Comment    `json:"comment.comments,omitempty"`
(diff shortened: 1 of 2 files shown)


andygrunwald:
Thank you

PR #28, opened 2016-07-28
Add epic data to returned issue fields

attack7:
* Adds a struct to contain the data about Epics in issues.  (I didn't include color - thought it was an unnecessary extra struct for an atlassian-specific UI case)
* A quick test to confirm that data is present


andygrunwald:
Thank you a  lot, Matt (@attack7)

PR #30, opened 2016-08-02
Adds a struct representing a version

attack7:
This adds a struct sufficient to hold the data points on each version returned when querying for a project.


andygrunwald:
Thank you :)

PR #31, opened 2016-08-16
Fixed a nil response when getting a session cookie

bah2830:
This fixes a panic error I was getting when a nil response was returned. This lets it fall through to the correct checks below that will return the appropriate error.


andygrunwald:
Thank you @bah2830

PR #33, opened 2016-09-06
IssueService.GetCustomFields: Read string values of complex type custom fields

lirao:
Attempt to retrieve the actual string value of a complex type custom field instead of blindly converting to string. 

e.g. a single choice select list would be retrieved as a string "map:[value:myvalue, id:29382 self:https://__]" instead of just "myvalue".


andygrunwald:
Thank you. Right now i don't have Access to a Computer. Only to my mobile
phone. Will Be back at
~18. September and than i will have a look. Just to let you know.

Thanks For this PR

Am Mittwoch, 7. September 2016 schrieb Coveralls :

andygrunwald:
This looks good!
Thank you!

PR #34, opened 2016-09-24
Added support for more jira elements

aviz:
Added group service and some jira service desk elements.

### DIFF
--- a/group.go
+++ b/group.go
@@ -1,0 +1,47 @@
+package jira
+
+import (
+	"fmt"
+)
+
+
+type GroupService struct {
+	client *Client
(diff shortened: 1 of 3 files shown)


andygrunwald:
Thanks @aviz and sorry for the delay.
I added the group functionality you provided in https://github.com/andygrunwald/go-jira/commit/8a4b1aca338f7f7a2b61bc6ad41db058c902dabd

The `IssueServiceDesk`-Code will not be implemented, because there you are referring to `customfield_10000`. This will be different in every JIRA installation. But this library has support for custom fields.

PR #36, opened 2016-09-27
Complete Session API. Adds CreateMeta api support

bidesh:
* Adds Logout and GetCurrentUser for session API. 
* Implements support for createMeta Endpoint.
* Changes to create issues by reading schema information returned from createMeta endpoint
 
The problem with Jira Issue schema is that the "fields" is like following:
```json
"fields": {
				"summary": {
					"required": true,
					"schema": {
						"type": "string",
						"system": "summary"
					},
					"name": "Summary",
					"hasDefaultValue": false,
					"operations": [
						"set"
					]
				}
}

```
Fields can have arbitrary keys like `customfield_xxxx`. Capturing these in the IssueFields already implemented is quite hard since the implementation of these fields can be diverse. So, IssueFields adds map structure to capture anything that is not already present in the struct fields. Custom MarshalJSON and UnmarshalJSON has been added. MarshalJSON moves everything from Unknowns to the same level as any other field. UnmarshalJSON puts all the fields with unknown json tags into `Unknown` map.

The choice for this is that when creating issues based on metadata, we do not have to depend on structs fields. This makes it alot more flexible and creating issues for arbitrary projects more convenient IMHO.

### DIFF
--- a/authentication.go
+++ b/authentication.go
@@ -133,6 +133,9 @@ func (s *AuthenticationService) GetCurrentUser() (*Session, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Error sending request to get user info : %s", err)
 	}
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("Getting user info failed with status : %d", resp.StatusCode)
+	}
 
 	defer resp.Body.Close()
 	ret := new(Session)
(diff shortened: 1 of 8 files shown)


andygrunwald:
Merged!

PR #41, opened 2016-10-07
Fixes the structs tags. Adds test to verify empty fields are ommitted

bidesh:
The structs tag were wrongly named as "struct" at some places. This caused unproper marshalling of IssueFields. This has been fixed and a unit test added to verify the tags which shouldn't be there isn't. 
@bigkraig this will fix the issue.


bigkraig:
👍 LGTM

andygrunwald:
Thank you

PR #42, opened 2016-10-10
NewRawRequest method to use a native io.Reader instead of marshalling

gm42:
When using custom fields it would be impossible to annotate the field name and/or create it dynamically at run time (it would involve reflection black magic and other weird tricks).

The most simple solution I could find is to simply emulate the native `http.NewRequest` parameters as per this PR.

### DIFF
--- a/jira.go
+++ b/jira.go
@@ -61,6 +61,35 @@ func NewClient(httpClient *http.Client, baseURL string) (*Client, error) {
 	return c, nil
 }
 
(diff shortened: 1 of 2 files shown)


gm42:
I have no idea why it shows coverage decreased...I just added a test too.

andygrunwald:
Thank you @gm42 and sorry for the delay.

gm42:
glad to be of help :blush: thanks

PR #47, opened 2016-11-01
Added Parent to issue fields

dontrebootme:
This PR adds support for creating sub tasks associated with a parent issue by name or id.
https://docs.atlassian.com/jira/REST/cloud/#api/2/issue-createIssue
```
Creating a sub-task is similar to creating a regular issue, with two important differences:
the issueType field must correspond to a sub-task issue type (you can use /issue/createmeta to discover sub-task issue types), and
you must provide a parent field in the issue create request containing the id or key of the parent issue
```

PR #48, opened 2016-11-01
Added example for subtask creation

dontrebootme:


andygrunwald:
This is the same like #47 which is already merged.

PR #51, opened 2016-11-22
Fix User.GET() method

pmoncadaisla:
According to https://docs.atlassian.com/jira/REST/cloud/#api/2/user-getUser username must be provided as a query string parameter

### DIFF
--- a/user.go
+++ b/user.go
@@ -31,7 +31,7 @@ type User struct {
 //
 // JIRA API docs: https://docs.atlassian.com/jira/REST/cloud/#api/2/user-getUser
 func (s *UserService) Get(username string) (*User, *Response, error) {
-	apiEndpoint := fmt.Sprintf("/rest/api/2/user/%s", username)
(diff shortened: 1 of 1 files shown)


javierprovecho:
@pmoncadaisla @andygrunwald test fails because mock http server is not defined correctly in https://github.com/andygrunwald/go-jira/blob/master/user_test.go#L12-L14

```go
...
	testMux.HandleFunc("/rest/api/2/user", func(w http.ResponseWriter, r *http.Request) {
		testMethod(t, r, "GET")
		testRequestURL(t, r, "/rest/api/2/user?username=fred")
...
```

It's funny because the object returned was already wrong. https://github.com/andygrunwald/go-jira/blob/master/user_test.go#L16 shows that `self` is following Atlasian docs.

andygrunwald:
Thank you @pmoncadaisla and @javierprovecho 
Merged and i fixed the unit tests.

PR #54, opened 2016-12-13
Make CustomFields return an interface instead of a string of whatever the field is

roganartu:
This allows dealing with dropdown boxes and checkboxes etc as they are returned as JSON objects.


roganartu:
Obsoleted by latest changes I hadn't pulled down from upstream, rejecting.

andygrunwald:
Glad that it works for you :)

Issue #45, opened 2016-10-26
Add changelog


dobegor:
It would be nice to have a way to provide "expand" parameter to get issue's changelog, i.e.
`http://your-jira-server/rest/api/2/issue/ISSUE-11?expand=changelog`

andygrunwald:
@dobegor True. This is not supported yet.
Right now i don`t have the time to implement it.

For the implementation i would choose the package `github.com/google/go-querystring/query`.
With this you are able to define "Option"-Structs with `url` tags like here: https://github.com/andygrunwald/go-gerrit/blob/master/projects.go#L169

Based on this struct you can implement an `addOptions` like here: https://github.com/andygrunwald/go-gerrit/blob/master/gerrit.go#L438

And here is an example how you would call this `addOptions` function: https://github.com/andygrunwald/go-gerrit/blob/eba8c6989a186784d5239df46ba5d166d7c17fdd/projects_tag.go#L22

Does this make sense to you?
Would you mind to do a PR?

andygrunwald:
Hey @dobegor,
@lycis was so kind and added ChangeLog functionality in #58
I hope this solved your request. If not feel free to add a new issue.

PR #58, opened 2017-01-11
Issue Changelog and Query Parameters for Issue.Get

lycis:
With this PR I added the following functionality:

* Retrieve the unmarshal the changelog of Issues when present
* Add a way to append query parameters (e.g. `...?expand=changelog`) to Issue.Get(...) requests

This would address issue #45 in a way. Merge or provide suggestions for improvement or decline as you like.


andygrunwald:
Thank you @lycis!
I applied your change a little bit different and added a unit test. But the functionality should be the same.
The commit was in https://github.com/andygrunwald/go-jira/commit/f764e89d33b6fd160efffc23be65701b4f21c932

Thank you!

PR #63, opened 2017-02-20
Fix relative path for ProjectService.Get

phille97:


phille97:
This is a problem if the base URL is something like example.com/jira/

phille97:
Example: https://play.golang.org/p/50KjcEZJmJ

andygrunwald:
Thank you!

Issue #60, opened 2017-01-27
Add support for issue deletion


niallo:
Thanks for writing this JIRA library. I noticed that issue deletion is not supported. It would be really nice to have this. Cheers.

andygrunwald:
Hey @niallo,

you are right. Good catch. This functionality is not implemented (yet).
But i think this is useful. I would be happy to accept a pull request for this.
Here are a few notes that might get you or someone else who is willing to contribute up to speed:

* The requested functionality is documented at [Delete Issue @ JIRA Docs](https://docs.atlassian.com/jira/REST/cloud/#api/2/issue-deleteIssue)
* A similar function (not for issues, but for "Boards") is implemented already in go-jira: [DeleteBoard @ `BoardService`](https://github.com/andygrunwald/go-jira/blob/a7647f3c7b3605a6fd178aba217adba6131eb4b8/board.go#L130)
* The code is (nearly) the same like the one in BoardService with the adjustments to the `IssueService` and the API-URL
* How a functionality for the `IssueService` will be implemented can be seen at [Get @ `IssueService`](https://github.com/andygrunwald/go-jira/blob/a7647f3c7b3605a6fd178aba217adba6131eb4b8/issue.go#L426)
* If you want to go even further a unit test for this is highly welcome!
* Such a unit test needs to be implemented in the [`issue_test.go`](./issue_test.go) file
* How a unit test for a DELETE call can look like is visible at [`TestBoardService_DeleteBoard`](https://github.com/andygrunwald/go-jira/blob/a7647f3c7b3605a6fd178aba217adba6131eb4b8/board_test.go#L136)

andygrunwald:
Was fixed in #65
Thanks go to @ggrpmnn

ggrpmnn:
:+1:

PR #65, opened 2017-02-24
Feature addition: issue delete

ggrpmnn:
I saw the Issue.Delete issue open (#60) and took a stab at adding the code. My first commit fixes several minor issues that atom's `gometalinter` plugin discovered, and the next commit adds the delete functionality and unit test. Let me know if anything can be improved or needs to be changed.

### DIFF
--- a/metaissue.go
+++ b/metaissue.go
@@ -37,7 +37,7 @@ type MetaIssueType struct {
 	IconUrl     string                `json:"iconurl,omitempty"`
 	Name        string                `json:"name,omitempty"`
 	Subtasks    bool                  `json:"subtask,omitempty"`
-	Expand      string                `json:"expand,omitempty"`
+	Expand      string                `json:",omitempty"`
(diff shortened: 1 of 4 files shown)


metaissue.go
```diff
@@ -37,7 +37,7 @@ type MetaIssueType struct {
 	IconUrl     string                `json:"iconurl,omitempty"`
 	Name        string                `json:"name,omitempty"`
 	Subtasks    bool                  `json:"subtask,omitempty"`
-	Expand      string                `json:"expand,omitempty"`
+	Expand      string                `json:",omitempty"`
```
andygrunwald: Why this was removed?

andygrunwald:
Thanks for this PR. It looks good. Only the one question in the metaissue.go is open

metaissue.go
ggrpmnn: It was causing a `go vet` issue because both Self and Expand had the JSON expand key.

andygrunwald:
Thank you :)

PR #68, opened 2017-04-13
adds support for selecting specific fields from the jira api. See htt…

wvell:
…ps://docs.atlassian.com/jira/REST/cloud/#api/2/search-search for details


andygrunwald:
Thank you

PR #71, opened 2017-04-27
Add the "option" field type for custom fields

antifuchs:
## What does this PR do?
JIRA supports SelectList field types, which are similar to but
represented not quite the same as Strings. This change allows go-jira
to create issues with custom fields on them that are SelectLists.

## Why?

I had to create an issue in a project that heavily templates issues with custom fields, and it's making use of the "SelectList" type a lot. This was the only way I could get it to work on our installation.

I hope you find this PR acceptable - happy to fix anything that's wrong with it (:

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -312,6 +312,12 @@ type TransitionPayload struct {
 	ID string `json:"id" structs:"id"`
 }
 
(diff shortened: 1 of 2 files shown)


andygrunwald:
Thank you :)

PR #75, opened 2017-06-15
Add Issue Update method

juicemia:
JIRA only allows issues to be created with certain fields set. This
means if a user needs to create an issue through the API with the
description set, they need to create the issue first, and then call the
API again to update it.

Without this commit, the way to do this through the package is by
calling IssueService.Create and then use the client's NewRequest method
to create a raw HTTP request for the Update. This isn't bad, but it can
be better.


andygrunwald:
Thank you @juicemia

PR #76, opened 2017-06-26
Fix bug in issue update

juicemia:
The issue was being found by ID, but there are cases where a user will
have the issue key but not the ID, whereas if the user has the ID then
it's highly likely they have the key, because they probably got the ID
from the API, which would also return the key.


andygrunwald:
Thank you

PR #79, opened 2017-08-03
add UpdateIssue method for jira 7.4.0

netxfly:


issue.go
```diff
@@ -608,6 +608,27 @@ func (s *IssueService) Update(issue *Issue) (*Issue, *Response, error) {
 	return &ret, resp, nil
 }
 
+// Update updates an issue from a JSON representation. The issue is found by key.
+//
+// https://docs.atlassian.com/jira/REST/7.4.0/#api/2/issue-editIssue
+func (s *IssueService) UpdateIssue(jiraId string, data map[string]interface{}) (*Response, error) {
+	apiEndpoint := fmt.Sprintf("rest/api/2/issue/%v", jiraId)
+	req, err := s.client.NewRequest("PUT", apiEndpoint, data)
+	fmt.Println(apiEndpoint, req, err)
```
andygrunwald: Please remove this side effect here. Every tool that uses the message would have an output on stdout/stderr with this.

issue.go
```diff
@@ -608,6 +608,27 @@ func (s *IssueService) Update(issue *Issue) (*Issue, *Response, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	resp, err := s.client.Do(req, nil)
+	fmt.Println(resp, err)
```
andygrunwald: Please remove this side effect here. Every tool that uses the message would have an output on stdout/stderr with this.

issue.go
```diff
@@ -608,6 +608,27 @@ func (s *IssueService) Update(issue *Issue) (*Issue, *Response, error) {
...
+	if err != nil {
+		return resp, err
+	}
+
+	// This is just to follow the rest of the API's convention of returning an issue.
```
andygrunwald: This comment can be removed, because we don't return the same issue.

andygrunwald:
Thank you @netxfly

PR #87, opened 2017-10-28
Add GoDoc examples

blachniet:
Converted examples in the README to GoDoc examples. You can [see what this will look like in GoDoc here](https://godoc.org/github.com/blachniet/go-jira#pkg-examples). I also added an example in which you can ignore certificate errors for test JIRA instances and those configured with self-signed certificates: [NewClient, Ignore Certificate Errors](https://godoc.org/github.com/blachniet/go-jira#example-NewClient--IgnoreCertificateErrors).

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -1,0 +1,130 @@
+package jira_test
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net/http"
+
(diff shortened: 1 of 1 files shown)


andygrunwald:
This is great. Thanks a lot @blachniet

Issue #55, opened 2016-12-14
Add support for pagination


stevegore:
The JIRA REST API applies pagination to a number of resources (e.g. users in group, issues returned from a search): https://docs.atlassian.com/jira/REST/server/#pagination

Ideally we would fetch the first page, and the fetch other pages concurrently.

Currently, it seems like this needs to be done manually with each implementation of the library.

andygrunwald:
Sorry for the late reply @stevegore.
You are right, sophisticated pagination support is not implemented (yet).
At the moment you can reach this feature by applying the `StartAt` in your `SearchOptions` in your application code.

This can have pro and cons.
Cons: Every app needs to implement this
Pro: You might not flood your JIRA instance with requests.

Your idea to request this concurrent is quite nice. Thank you for this.

If someone would implement this and apply a PR, i would be happy to merge it.
Right now i don't have time to implement it myself. Sorry.
Would you mind to give it a try and add this feature?

ghostsquad:
I would strong advice against automating any sort of "fetch all" functionality, as with the wrong query, this could easily timeout, or fetch more items that can be managed. Instead, I think we could implement a `PaginatedResult` type object, that simply makes it very easy to request the next page.

Some ideas for reference:
https://github.com/qiangxue/golang-restful-starter-kit/blob/master/util/paginated_list.go

I didn't find much in the way of lazy evaluation in Go, but I suspect it's as simple as returning a function to the caller, which when invoked could fetch the next page.

```go
func (p *PaginatedResult) Next() func() *PaginatedResult {
   ...
}

```

andygrunwald:
After some time, I am against my own comment from 2017.
Pagination should be handled more explicitly.

Happy to accept PRs in this direction.

fdcds:
> At the moment you can reach this feature by applying the `StartAt` in your `SearchOptions` in your application code.

With `GetIssuesForSprint` I ran into the issue that I do not see a way to specify `StartAt` (#302).

eparis:
Since I ended up here I figured I'd post this for the next person implementing it in your code.
```
func GetIssues(client *jira.Client, searchString string) ([]jira.Issue, error) {
        last := 0
        var issues []jira.Issue = nil
        for {
                opt := &jira.SearchOptions{
                        MaxResults: 100,
                        StartAt:    last,
                }

                chunk, resp, err := client.Issue.Search(searchString, opt)
                if err != nil {
                        return nil, err
                }

                total := resp.Total
[...]

PR #88, opened 2017-11-01
Add SearchPages

guywithnose:
This starts to implement the changes for #55.

It doesn't run concurrently yet, but I believe this implementation can be made concurrent later.  Implementing concurrency is a bit more complicated and I have a few things to consider.

I think we should be careful to not call `f` from more than one goroutine.  Otherwise, the user will need to ensure they are processing the results in a thread safe manner.  I also think it is important to maintain the order the issues are returned.

With these two constraints, the concurrent implementation will certainly add quite a bit of complexity, so I hope this is good for now.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -723,6 +723,46 @@ func (s *IssueService) Search(jql string, options *SearchOptions) ([]Issue, *Res
 	return v.Issues, resp, err
 }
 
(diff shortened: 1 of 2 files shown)


andygrunwald:
Fine for me :) Thank you!

PR #89, opened 2017-11-02
Adding jira.Error to better wrap jira-specific errors

rbriski:
This change adds a little more functionality around the error returned from Jira services (e.g. `IssueService`, etc).  It's backward compatible.

Jira errors currently require an extra step of checking the response to understand what went wrong.  Here is an example:
```
Request failed. Please analyze the request body for more details. Status code: 404
```

This change wraps the HTTP error in a jira.Error struct and parses the jira error as well.  The short error now looks something like this:
```
Error: Issue does not exist or you do not have permission to see it.: Request failed. Please analyze the request body for more details. Status code: 404
```

There is also an option to check the long form of the error if you want with code like this:
```go
if err, ok := err.(*jira.Error); ok {
  fmt.Println(err.LongError())
} else {
  fmt.Println(err)	
}
```

The result looks like this:
```
Original:
Request failed. Please analyze the request body for more details. Status code: 404
Messages:
 - Issue does not exist or you do not have permission to see it.
```

Initially, I have only changed the `Get` function in the `IssueService`.  If you think it's useful, I can add it more widely.


rbriski:
Older go versions are failing because of a dependency error.  I'm not sure how this project handles dependencies so I can't specify the `httptest` version that I'm using.

rbriski:
I was using `httptest` directly.  I didn't realize there was a `setup()` procedure that other tests use to mock the response.  Once I switched to the custom `testClient` the tests pass.

rbriski:
I'm going to close and reopen another branch because of the weird fork conflict.

andygrunwald:
@rbriski Thanks a lot. Useful contribution.
As far as i can see this don't affect all endpoints, right?
If you have the time to add the other endpoints as well, feel free :)

PR #90, opened 2017-11-03
DoTransitionWithPayload should return resp on failure

pavel-kolesnikov:


andygrunwald:
@pavel-kolesnikov Thank you. Good addition.
Would you mind to rebase your patch? We had a merge in between.

Further more i think https://github.com/andygrunwald/go-jira/blob/master/error.go#L21 needs to be adjusted if err is nil, that `NewJiraError` returns nil as well.

pavel-kolesnikov:
will rethink the patch ;)

Issue #91, opened 2017-11-14
Runtime error when authentication error occurs


jeynesrya:
When I use an invalid host or the authentication details are incorrect, the errors printed back to the console are not best helpful when it fails to get an issue. See below for output currently:

```
_panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6de982]

goroutine 1 [running]:
github.com/andygrunwald/go-jira.NewJiraError(0x0, 0x8a9160, 0xc4201182a0, 0x0, 0x0)
	.../go/src/github.com/andygrunwald/go-jira/error.go:22 +0x52
github.com/andygrunwald/go-jira.(*IssueService).Get(0xc42000e038, 0x779d93, 0x8, 0x0, 0x0, 0x0, 0x0, 0x0)
	.../go/src/github.com/andygrunwald/go-jira/issue.go:499 +0x512
main.main()
	.../golang/src/github.com/jeynesrya/trial-go/jira.go:15 +0x19b_
```

Seems as though an error is occurring whilst the error is attempting to be built. I have to dig down into breakpointed code in order to find what the actual error was. For example, I have to dig down in the httpError variable in order to find that my issue was actually "no such host" but this is never returned back to the console.

rbriski:
That looks like an error against the code I added. I'll look at it tonight.  Do you mind providing more detail around how you call the code so I can get the rest case right?

rbriski:
I can't replicate this by changing the host to something non-existent or messing up auth.  However, if I turn off my connection, I can get this error to happen.  Working on it.

rbriski:
Turns out, the custom `JiraError` was eating the error from the http connection.  The PR above checks and wraps the http error if there is no response.

andygrunwald:
Thanks @rbriski for taking care.
@jeynesrya can you test the new version (master) and confirm that if #92 fixes your problem?

rbriski:
Based on the issue and the testing I did, I believe the PR closed this.  You can probably close this issue.

andygrunwald:
Great. Thanks to all!

jeynesrya:
Confirmed, sorry for the delay

PR #92, opened 2017-11-15
Fixed null pointer when no response is received (#91)

rbriski:
* JiraError can now handle nil responses
* Better error handling when the response is not JSON
* Closes issue #91


andygrunwald:
Thanks a lot @rbriski

PR #94, opened 2017-11-27
Finding users by email, username or name

manute:
Hi ,

I'm using this lib and it's very nice. I was trying to find a user by email and I realized that it hasn't got the method for it. So I have created this PR to find users by email, username or name.

It retrieves a list of users and it is based on this https://docs.atlassian.com/jira/REST/cloud/#api/2/user-findUsers

### DIFF
--- a/user.go
+++ b/user.go
@@ -72,3 +72,22 @@ func (s *UserService) Create(user *User) (*User, *Response, error) {
 	}
 	return responseUser, resp, nil
 }
+
+// Finds user info from JIRA:
+//  it can find users by email,username or name
+//
+// JIRA API docs: https://docs.atlassian.com/jira/REST/cloud/#api/2/user-findUsers
+func (s *UserService) Find(property string) ([]User, *Response, error) {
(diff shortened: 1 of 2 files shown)


rbriski:
Not necessarily a review of the code, but if you use the `NewJiraError` function on the returned error like the link below, you'll get more actionable error messages:

https://github.com/andygrunwald/go-jira/blob/master/issue.go#L517

manute:
I have followed the same workflow is used in all the user's methods , so I think that change would be in other PR(for changing all possible methods). 
Otherwise I will mix here methods with that response and other without that response. 
What do you think?

rbriski:
Mmm, good point.  I must have missed the user method return JSON.  I thought I caught them all.  Sounds fine to me although it's obviously up to @andygrunwald

rbriski:
LGTM

andygrunwald:
Thanks @manute for the PR and thanks @rbriski for the review.
In https://github.com/andygrunwald/go-jira/commit/5ac13b726c0a3510c801227997a781cb8f8e3eac i applied the `NewJiraError` function

PR #95, opened 2017-11-27
Oauth2 Support

### LINKED ISSUE #93: support for oauth2 used by Jira Add-Ons
In order to support REST API calls from Jira Add-Ons created through the Atlassian Connect framework, we need to: 

1. create a JSON Web Token (JWT) assertion
2. use the JWT assertion + secret to get an access token
3. include the access token in the request headers

I've got a working standalone example and I'll create a PR in a bit.

shdunning:
Fixes #93 

The basic gist would be to do the following from the client code:

1. Get / Set new oauth2 access token:
```go
jiraClient, _ := jira.NewClient(nil, instanceURL)
token, err := jiraClient.AuthenticationService.GetOauth2AccessToken(oauthId, userKey, scopes, secret)
// error check and, optionally, cache token if need-be
cacheToken(token, instanceURL, scopes)
```

2. Use a stored access token:
```go
jiraClient, _ := jira.NewClient(nil, instanceURL)
// retrieve valid access token from cache
token := getCachedToken(instanceURL, scopes)
jiraClient.AuthenticationService.SetOauth(token)
```

Todo:

- [x] finish tests
- [x] update README
- [x] address code review comments


authentication.go
```diff
@@ -3,15 +3,22 @@ package jira
...
+	"time"
 )
 
 const (
 	// HTTP Basic Authentication
 	authTypeBasic = 1
 	// HTTP Session Authentication
 	authTypeSession = 2
+	// Oauth Authentication
+	authTypeOauth = 3
```
rbriski: I'd call out the oauth version in case another oauth1 is added.  For example, `authTypeOauth2`

authentication.go
```diff
@@ -89,6 +106,13 @@ func (s *AuthenticationService) AcquireSessionCookie(username, password string)
 	return true, nil
 }
 
+// Sets the access token on the Authentication Service (helpful in event access tokens are cached by the client code).
+// Use GetOauth2AccessToken to get a new access token and set authType to Oauth
+func (s *AuthenticationService) SetOauth(accessToken string) {
```
rbriski: Same with version number, or pass version number to the method:

*Method name*
`SetOauth2`

*Argument*
`SetOauth(accessToken string, version int)`

authentication.go
```diff
@@ -178,3 +204,75 @@ func (s *AuthenticationService) GetCurrentUser() (*Session, error) {
...
+// Attempts to retrieve an Oauth2 access token from the Jira Auth Server. Returns the details of the access token
+// in a JiraAccessToken struct so that it can be cached by the client application if desired (oauth2 tokens expire
+// after 15 min by default); else, this can be called multiple times to retrieve a new access token.
+//
+// oauthClientId, userKey, and sharedSecret are returned by Jira during the installation of the Add-on
+// scope is the scope to place on the AccessToken returned (e.g., "READ", "READ WRITE", "READ ACT_AS_USER")
+// See: https://developer.atlassian.com/cloud/jira/software/oauth-2-jwt-bearer-token-authorization-grant-type
+func (s *AuthenticationService) GetOauth2AccessToken(oauthClientId, userKey, scope string, sharedSecret []byte) (*JiraAccessToken, error) {
+	// create JSON Web Token
+	jiraAuthURL := "https://auth.atlassian.io"
```
rbriski: This seems like it should be:
1. A constant at the top of the `AuthenticationService`
2. A config variable with a default

I'm not sure how many oauth2 services that jira will possibly expect.  That would inform which one you'd pick, I think.

authentication.go
```diff
@@ -178,3 +204,75 @@ func (s *AuthenticationService) GetCurrentUser() (*Session, error) {
...
+		"sub": fmt.Sprintf("urn:atlassian:connect:userkey:%s", userKey),
+		"aud": jiraAuthURL,
+		"iat": time.Now().Unix(),
+		"exp": time.Now().Unix() + 10, // expires in 10 seconds
+	})
+
+	// sign the jwt using the signing method and the shared secret
+	if signingStr, err := jwt.SigningString(); err != nil {
+		return "", err
+	} else {
```
rbriski: Not sure this needs an `else` since you've called `return` early on the previous error.

authentication.go
```diff
@@ -178,3 +204,75 @@ func (s *AuthenticationService) GetCurrentUser() (*Session, error) {
...
+		"tnt": instanceURL,
+		"sub": fmt.Sprintf("urn:atlassian:connect:userkey:%s", userKey),
+		"aud": jiraAuthURL,
+		"iat": time.Now().Unix(),
+		"exp": time.Now().Unix() + 10, // expires in 10 seconds
+	})
+
+	// sign the jwt using the signing method and the shared secret
+	if signingStr, err := jwt.SigningString(); err != nil {
+		return "", err
```
rbriski: Up to you, but I tend to wrap errors to give more context, if it's useful:
```go
return "", errors.Wrap(err, "Could not sign string")
```

authentication.go
rbriski: After looking at the rest of the PR, it looks like this whole thing is very closely tied to the JIRA-specific oauth2 service, so probably option 1 above.

authentication.go
```diff
@@ -178,3 +204,75 @@ func (s *AuthenticationService) GetCurrentUser() (*Session, error) {
...
+//
+// oauthClientId, userKey, and sharedSecret are returned by Jira during the installation of the Add-on
+// scope is the scope to place on the AccessToken returned (e.g., "READ", "READ WRITE", "READ ACT_AS_USER")
+// See: https://developer.atlassian.com/cloud/jira/software/oauth-2-jwt-bearer-token-authorization-grant-type
+func (s *AuthenticationService) GetOauth2AccessToken(oauthClientId, userKey, scope string, sharedSecret []byte) (*JiraAccessToken, error) {
+	// create JSON Web Token
+	jiraAuthURL := "https://auth.atlassian.io"
+	jwtStr, err := createJWT(oauthClientId, s.client.baseURL.String(), jiraAuthURL, userKey, sharedSecret)
+	if err != nil {
+		return nil, fmt.Errorf("Error creating JSON Web Token : %s", err)
```
rbriski: `Wrap` works better for this:
```golang
return nil, errors.Wrap(err, "Error creating JSON Web token")
```

rbriski:
If @andygrunwald is OK with adding `oauth2` into the `AuthenticationService` I can add the `oauth1` work that I did to clean up the earlier gist.

authentication.go
shdunning: `signingStr` loses scope outside of the if / else block and I only need it for the subsequent call.

authentication.go
```diff
@@ -143,7 +174,7 @@ func (s *AuthenticationService) Logout() error {
 // JIRA API docs: https://docs.atlassian.com/jira/REST/latest/#auth/1/session
 func (s *AuthenticationService) GetCurrentUser() (*Session, error) {
 	if s == nil {
-		return nil, fmt.Errorf("AUthenticaiton Service is not instantiated")
+		return nil, fmt.Errorf("Authenticaiton Service is not instantiated")
```
rbriski: Typo: Authentication

shdunning:
@andygrunwald any update here?

andygrunwald:
Discussion ongoing inn #93

andygrunwald:
In https://github.com/andygrunwald/go-jira/pull/112 @rbriski moved the existing auth options into a RoundTripper.
@shdunning Do you think this would also make sense here?

shdunning:
@andygrunwald I think so. If I'm understanding correctly we would introduce another type that implements the `RoundTripper` interface (say, `Oauth2Transport`) and add a method similar to: https://github.com/andygrunwald/go-jira/blob/master/jira.go#L345.

The 2 steps specific to Oauth2 are:
1. JWT assertion construction (during token request)
2. Specify the Bearer Token in the headers (during API request, after the token has been granted).

#2. is pretty standard for oauth2 but #1. has some [Jira-specific steps](https://developer.atlassian.com/cloud/jira/software/oauth-2-jwt-bearer-token-authorization-grant-type/). Were you and @rbriski envisioning that JWT construction would be up to the client app, or would we have so helpers in `go-jira` to help construct the right stuff?

rbriski:
I'm close to having an example project up for JWT authentication.  I hope to commit the first draft of the code before the end of this weekend.  JWT authentication is just for JIRA add-ons read, write capability.  The Bearer token, needed for ACT_AS_USER capability, I plan on doing next weekend.  Stay tuned.  I'll update this when I push the first bit of code.

rbriski:
OK, I'm not finished but I have enough to show how authentication works with `RoundTripper` and `go-jira`.  The repo is here:
https://github.com/rbriski/atlassian-jwt

I have an example that can be run and loaded as an add-on here:
https://github.com/rbriski/atlassian-jwt/blob/master/examples/jwt/main.go

Like I said, it's quite early so not ready for use.  Eventually, I hope to use it as a drop in for JWT (and possible oauth2) JIRA auth on `go-jira`

I'm closing this PR and moving the conversation over the relevant issue (#93 )

PR #100, opened 2018-01-10
Change type to Time for issue time fields

omira-sch:
Set the type to 'Time' for time-based fields in 'IssueFields'.
* Make it easier to perform time operations, like get the diff between two dates.
* Be consistent with 'WorklogRecords' fields.
Some fields are optional, so it checks for 'null' to do not raise an error.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -101,12 +101,12 @@ type IssueFields struct {
 	Project              Project       `json:"project,omitempty" structs:"project,omitempty"`
 	Resolution           *Resolution   `json:"resolution,omitempty" structs:"resolution,omitempty"`
 	Priority             *Priority     `json:"priority,omitempty" structs:"priority,omitempty"`
-	Resolutiondate       string        `json:"resolutiondate,omitempty" structs:"resolutiondate,omitempty"`
-	Created              string        `json:"created,omitempty" structs:"created,omitempty"`
-	Duedate              string        `json:"duedate,omitempty" structs:"duedate,omitempty"`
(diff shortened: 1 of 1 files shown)


andygrunwald:
Thank you

takahito-tejima:
wondering should other fields such as ChangelogHistory.Created and Comment.Updated have the Time type as well for consistency?

btw go-jira is super useful, I love it! thanks for sharing a great library.

takahito-tejima:
And I just realized a problem possibly related to this change.
When I put "Duedate" to an issue such as "2018-02-01", jiraClient.Issue.Search raises an error

> parsing time ""2018-02-01"" as ""2006-01-02T15:04:05.999-0700"": cannot parse """ as "T": http: read on closed response body

this is because Duedate can be set by user and not automatically filled unlike other fields

omira-sch:
@takahito-tejima take a look at #102

andygrunwald:
@takahito-tejima 

> wondering should other fields such as ChangelogHistory.Created and Comment.Updated have the Time type as well for consistency?

This would make sense, yes. Would you mind to prepare the change and send a new PR?

> btw go-jira is super useful, I love it! thanks for sharing a great library.

You are welcome! This library is not my work alone. A lot of contributors pushed _very useful_  changes in.

> [...] this is because Duedate can be set by user and not automatically filled unlike other fields

#102 is merged.
Thanks to @Xjs for the PR and for @omira-sch for pointing it to the right PR.

PR #101, opened 2018-01-10
New TransitionPayloadFields to set fields on transitions

omira-sch:
* New 'TransitionPayloadFields' to set field values when executing transitions (e.g. the built-in 'Resolution' field which is always required to resolve issues).
* Add 'To' field to transitions to be able to work with custom projects. This way you can check 'To.StatusCategory' to do automatic transitions in workflows with non-standard transition names.


andygrunwald:
Thank you :)

PR #102, opened 2018-01-16
Use custom "Date" type for Duedate field

Xjs:
Jira doesn't use its default time format
("2006-01-02T15:04:05.999-0700") for the duedate field, but instead a
shorter representation that contains only the year, month and day
fields, i. e. "2006-01-02". This commit adds a new type "Date" that
is unmarshalled using the short format.


andygrunwald:
Thanks @Xjs 
Can you add a unit test for this also?

Xjs:
Like [this](https://github.com/andygrunwald/go-jira/pull/102/commits/1a2c703f7f508af47f2dc1d79aa028eb0beea261)? Sadly, GitHub doesn't highlight the changed characters, only the lines. The only change is the addition of a "duedate" field near the end of each of the JSON lines, after the "updated" field.

 (I'm not entirely sure if I caught everything, but I think these are the only tests where the Duedate field is populated from JSON.)

andygrunwald:
Thanks a lot @Xjs 
Sorry for the late merge. Had some priorities in my RL.

PR #104, opened 2018-01-31
adds ability to list projects with query params

shdunning:
The list projects API endpoint in Jira accepts the standard GetQueryOptions params (like `expand=`). This PR adds an extension to the ProjectService to pass these params through and refactors the existing `GetList` top use the new method. 

@rbriski and @andygrunwald let me know what you guys think.

### DIFF
--- a/project.go
+++ b/project.go
@@ -2,6 +2,8 @@ package jira
 
 import (
 	"fmt"
+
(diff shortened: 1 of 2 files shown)


andygrunwald:
Thank you.

PR #107, opened 2018-02-17
Added method & tests for logging time against an issue

robiball:


andygrunwald:
Thank you @robiball :)

PR #109, opened 2018-02-17
Add dep as a dependency manager

rbriski:
`dep` is on it's way to being included in the general Go tooling.  I think it's safe to say it's the winner of the dependency management tools discussion.  This PR sets it up for use in `go-jira`.  

I did not include `vendor`.  I can if it's very important that the project has guaranteed reproducible builds.  It seemed like unnecessary complexity at this time.


andygrunwald:
Thanks @rbriski!

This makes sense at all, but i am not 100% sure if this make sense for a library.
Because often go-jira is used as a lib in other projects (top level apps). And the top level apps should use dep to version go-jira.

Is dep in the top level app respecting the version pinning of go-jira itself?

rbriski:
I was thinking the same thing which is why I put it under the section on
contributions.

Dep will also respect versioning as long as the project uses semver.

The idea was part of a larger change to make it easier to contribute. Right
now it's a little hard to replicate bugs because to you have to create a
"parent" project that uses go-jira

andygrunwald:
After reading https://github.com/golang/dep/issues/208 it should be good to go ;)

andygrunwald:
Thank you.

PR #110, opened 2018-02-20
Update struct tags in IssueFields

acidbee:
Added 'omitempty` to summary and type fields in IssueFields. This will allow the IssueFields object to be used for updates and as payloads for transitions.


andygrunwald:
Thank you @acidbee

PR #111, opened 2018-02-23
Removed the / from the URL's to ensure no // in the request

jleeh:


andygrunwald:
Hey @jleeh,
any reason why you closed this immediately?

jleeh:
Hey @andygrunwald,

The PR's base was meant to be our forked version rather than merged back into here. 

I think the issue we were seeing was due to changes in our forked version, but our load-balancers for JIRA was throwing the request out because it was containing two slashes.

andygrunwald:
Got it.
What did you changed in your fork version?
Does it make sense to get this into upstream?

PR #112, opened 2018-02-24
Moving to RoundTripper for authentication

rbriski:
This is the first of a few PRs to move away from authentication directly in the library.  This PR handles basic authentication and cookie authentication.  I've updated the `README`, added an `examples` directory and put in the basic auth code.

There will be an issue updating the AuthenticationService since it handles auth instead of just reading from the endpoint.  It's also tied deeply into the `NewRequest` functionality of the client.  I'm still thinking about how best to do this in a backward compatible way.

I also still have to add tests for cookie authentication. 

Most of the `AuthenticationService` has now been deprecated.

This is ready to merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,35 +80,29 @@ func main() {
 }
 ```
 
-### Authenticate with jira
+### Authentication
 
-Some actions require an authenticated user.
+The `go-jira` library does not handle most authentication directly.  Instead, authentication should be handled within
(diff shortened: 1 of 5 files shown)


andygrunwald:
This looks really good to me. Thanks for all the work. I think this is a good basement for the OAuth change also

PR #113, opened 2018-02-28
Extend GetCreateMeta method to retrieve meta data for all projects

nguyen970:
Changed GetCreateMeta method to not require a projectKey. When a user passes in an empty string, it will retrieve the metadata for all projects.

Also changed "Subtasks" to "Subtask" to match Jira's data structure.

### DIFF
--- a/metaissue.go
+++ b/metaissue.go
@@ -36,7 +36,7 @@ type MetaIssueType struct {
 	Description string                `json:"description,omitempty"`
 	IconUrl     string                `json:"iconurl,omitempty"`
 	Name        string                `json:"name,omitempty"`
-	Subtasks    bool                  `json:"subtask,omitempty"`
+	Subtask     bool                  `json:"subtask,omitempty"`
 	Expand      string                `json:"expand,omitempty"`
 	Fields      tcontainer.MarshalMap `json:"fields,omitempty"`
 }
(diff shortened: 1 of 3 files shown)


metaissue.go
```diff
@@ -36,15 +36,22 @@ type MetaIssueType struct {
 	Description string                `json:"description,omitempty"`
 	IconUrl     string                `json:"iconurl,omitempty"`
 	Name        string                `json:"name,omitempty"`
-	Subtasks    bool                  `json:"subtask,omitempty"`
+	Subtask     bool                  `json:"subtask,omitempty"`
```
rbriski: Why did you change this?

metaissue.go
```diff
@@ -36,15 +36,22 @@ type MetaIssueType struct {
...
 	Expand      string                `json:"expand,omitempty"`
 	Fields      tcontainer.MarshalMap `json:"fields,omitempty"`
 }
 
 // GetCreateMeta makes the api call to get the meta information required to create a ticket
 func (s *IssueService) GetCreateMeta(projectkey string) (*CreateMetaInfo, *Response, error) {
 
-	apiEndpoint := fmt.Sprintf("rest/api/2/issue/createmeta?projectKeys=%s&expand=projects.issuetypes.fields", projectkey)
+	apiEndpoint := ""
```
rbriski: Please use `net/url` for modifying URLs, including query parameters.

metaissue.go
```diff
@@ -36,15 +36,22 @@ type MetaIssueType struct {
...
+	if len(projectkey) > 0 {
+		apiEndpoint = fmt.Sprintf("rest/api/2/issue/createmeta?projectKeys=%s&expand=projects.issuetypes.fields", projectkey)
+	}else {
+		apiEndpoint = fmt.Sprintf("rest/api/2/issue/createmeta?expand=projects.issuetypes.fields")
+	}
+
+	fmt.Println(apiEndpoint)
```
rbriski: Please remove this.

metaissue.go
nguyen970: I changed Subtasks to Subtask to match the jira data model. In addition, it is a boolean to check if the Issue Type is a subtask...so the 's' makes it confusing as it is not subtasks itself.

rbriski: But that will break this library for everyone else that's ever used it.

nguyen970:
I took the same approach as @shdunning with extending the GetList() function in projects. His [commit](https://github.com/andygrunwald/go-jira/commit/2267024e0373d8f74858de437e5b92f8a597be14).

PR #114, opened 2018-03-01
Cleaning a bunch of go vet issues

rbriski:
`go vet` enforces Go standards.  There were a couple places where the code didn't comply.  This fixes all of those except about 4 which would break the public API of the library.


andygrunwald:
It is about the small improvements :)
Thanks a lot!

PR #115, opened 2018-03-05
Fix GetQueryOptions in issue.go to use projectKeys instead of projectKey

Hoid:
Intended to fix GetQueryOptions to use projectKeys as its json key for ProjectKeys instead of projectKey. All other changes are made by the standard Go linter.


rbriski:
Won't this break backward compatibility? If someone is using the old field name their code will break.

Hoid:
I would think that the only thing affected would be (1) building their app, which would fail until they fixed the issue, and (2) if the user is attempting to use this projectKeys field, which already won't work. What could we do otherwise to fix this bug that wouldn't hurt backwards compatibility?

rbriski:
Good point and I just checked, this was inserted 4 days ago.  

LGTM 👍

PR #121, opened 2018-03-16
Add support for ValidateQuery param in SearchAPI

ishugupta92:


rbriski:
What happens if `Search` is called with `SearchOptions` and `ValidateQuery` isn't present?

ishugupta92:
The param remains blank like the other params in url and JIRA API uses the default value of `strict` which it would have used if the param had not been present at all.

rbriski:
Tested it.  It does default to strict if the field is blank.  Next time please include that test.  It'll let me merge the change faster.

Thanks!

ishugupta92:
I will make sure to do that next time. Thanks.

PR #123, opened 2018-04-25
Let the BoardService.GetAllSprints() return all the sprints.

pasinskim:
GetAllSprints() method was returning only the last 50 sprints by
default. This is adding a pagination to the
GET /rest/agile/1.0/board/{boardId}/sprint request and is iterating
over the responses until the `isLast` flag indicating that we've
fetched all the sprints is set.

### DIFF
--- a/board.go
+++ b/board.go
@@ -44,9 +44,23 @@ type BoardListOptions struct {
 	SearchOptions
 }
 
-// Wrapper struct for search result
-type sprintsResult struct {
-	Sprints []Sprint `json:"values" structs:"values"`
+// SprintListOptions specifies the optional parameters to the BoardService.GetAllSprints
+type SprintListOptions struct {
+	// State filters results to sprints in specified states.
+	// Valid values:  future, active, closed.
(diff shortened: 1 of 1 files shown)


rbriski:
Here's an example of how a person would use the pagination feature on `Search`:

```go
       opts := &jira.SearchOptions{
		StartAt:    0,
		MaxResults: 50,
	}
	jql := `some jql`

	var issueList []jira.Issue
	for {
		issues, resp, err := c.Issue.Search(jql, opts)
		if err != nil {
			return err		}

		issueList = append(issueList, issues...)
		if resp.StartAt+resp.MaxResults >= resp.Total {
			break
		}

		opts.StartAt += opts.MaxResults
	}
```

pasinskim:
@rbriski makes a lot of sense, and I will update the PR. Nevertheless, did you consider renaming the methods? `GetAllSprints` indicates that ALL the sprints will be retrieved, which is not true in fact. There are other similar methods as well (like `GetAllBoards` for example).

rbriski:
Resolved in #141.

PR #130, opened 2018-05-09
Some guards around url structures to help avoid bugs

### LINKED ISSUE #129: BUG: requests fail if baseUrl contains extra path
For Jira Server instances the base urls can contain extra path params (like https://example.com/jira) and `NewRequest` calls end up losing the last path param due to the way the request urls are built using `ResolveReference`:

https://github.com/andygrunwald/go-jira/blob/master/jira.go#L82
https://github.com/andygrunwald/go-jira/blob/master/jira.go#L119
https://github.com/andygrunwald/go-jira/blob/master/jira.go#L187

I *think* it should be safe to assume that the base urls are understood by developer and we should be keeping those intact and appending the relative url strings -- something like:

```
	rel, err := url.Parse(urlStr)
	if err != nil {
		return nil, err
	}

	u := path.Join(c.baseURL.String(), url.String())
```

The above will handle with or without preceding slashes in the relative url but won't handle crazy cases that [`ResolveReference` takes care of](https://golang.org/pkg/net/url/#example_URL_ResolveReference) -- not sure that all that defensive programming is necessary, tbh.

I'll submit a proposed bugfix today and see what you think.

shdunning:
Fixes #129 

2 changes:
1. add trailing `/` to baseURL if one wasn't specified
2. trim preceding `/` from url fragments in calls to `NewRequest` if fragment is not an absolute URL

This change helps to ensure full urls are built properly when `ResolveReference` is called from the `c.baseURL`.

### DIFF
--- a/jira.go
+++ b/jira.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/google/go-querystring/query"
 	"github.com/pkg/errors"
+	"strings"
 )
 
 // A Client manages communication with the JIRA API.
@@ -48,6 +49,11 @@ func NewClient(httpClient *http.Client, baseURL string) (*Client, error) {
 		httpClient = http.DefaultClient
 	}
 
+	// ensure the baseURL contains a trailing slash so that all paths are preserved in later calls
+	if !strings.HasSuffix(baseURL, "/") {
+		baseURL += "/"
+	}
+
 	parsedBaseURL, err := url.Parse(baseURL)
 	if err != nil {
 		return nil, err
@@ -71,12 +77,15 @@ func NewClient(httpClient *http.Client, baseURL string) (*Client, error) {
 
 // NewRawRequest creates an API request.
 // A relative URL can be provided in urlStr, in which case it is resolved relative to the baseURL of the Client.
-// Relative URLs should always be specified without a preceding slash.
 // Allows using an optional native io.Reader for sourcing the request body.
 func (c *Client) NewRawRequest(method, urlStr string, body io.Reader) (*http.Request, error) {
 	rel, err := url.Parse(urlStr)
 	if err != nil {
 		return nil, err
+	}
+	// Relative URLs should be specified without a preceding slash since baseURL will have the trailing slash
+	if !rel.IsAbs() {
+		rel.Path = strings.TrimLeft(rel.Path, "/")
 	}
 
 	u := c.baseURL.ResolveReference(rel)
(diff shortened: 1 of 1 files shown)


jira.go
```diff
@@ -71,13 +77,16 @@ func NewClient(httpClient *http.Client, baseURL string) (*Client, error) {
...
 // A relative URL can be provided in urlStr, in which case it is resolved relative to the baseURL of the Client.
-// Relative URLs should always be specified without a preceding slash.
 // Allows using an optional native io.Reader for sourcing the request body.
 func (c *Client) NewRawRequest(method, urlStr string, body io.Reader) (*http.Request, error) {
 	rel, err := url.Parse(urlStr)
 	if err != nil {
 		return nil, err
 	}
+	// Relative URLs should be specified without a preceding slash since baseURL will have the trailing slash
+	if !rel.IsAbs() {
```
rbriski: I don't think this is necessary.  Relative paths are defined within the library.  This is under our control.

rbriski: Same with the rest of the `LeftTrim` calls

jira.go
```diff
@@ -48,6 +49,11 @@ func NewClient(httpClient *http.Client, baseURL string) (*Client, error) {
 		httpClient = http.DefaultClient
 	}
 
+	// ensure the baseURL contains a trailing slash so that all paths are preserved in later calls
+	if !strings.HasSuffix(baseURL, "/") {
```
rbriski: Maybe this is just me, but I think I'd rather return an error.  I'm a little hesitant to assume intent.  Maybe some upstream library is stripping is or ruining the data passed in.  I think it'd be more helpful to explain the exact problem and where it occurred.

jira.go
shdunning: Except if clients call `c.NewRequest()` and use the returned request directly in `c.Do()` -- we do this in our usage a few places where we need to make API calls against API v1 (webhook APIs are still there). 

The way that we're constructing the full url does not preclude a client call passing in an absolute URL path here (i.e., it just works as expected).

shdunning: I guess **technically** a call to `TrimLeft` on an absolute url is a no-op so we could remove the check if we want to keep the PR...

jira.go
shdunning: @rbriski it does not produce an error at the time of initialization. When a request is made, the jira instance returns a 404 resource not found because the final url constructed from `NewRequest` can be wrong if the trailing `/` is omitted. This PR is help users not make the same mistake we did.

PR #131, opened 2018-05-12
Add new endpoint for getting assigned permission scheme for the project

b4b4r07:
## WHAY

Same as title

## WHY

This endpoint is not implemented now. So I'd like to add this.


rbriski:
This looks good. Can you add a positive test as well? Right now you just have a failing test.

andygrunwald:
LGTM as well. A positive test case would be appreciated before merging.
Thanks for your contribution!

b4b4r07:
Fixed! Could you review again?

PR #132, opened 2018-05-15
fix probable unexpected gzip issue

hanjm:
"Request failed. Please analyze the request body for more details. Status code: 401: Could not parse JSON: invalid character '\\x1f' looking for beginning of value"

### DIFF
--- a/jira.go
+++ b/jira.go
@@ -353,7 +353,10 @@ func (t *CookieAuthTransport) RoundTrip(req *http.Request) (*http.Response, erro
 			return nil, errors.Wrap(err, "cookieauth: no session object has been set")
 		}
 	}
-
+	// fix probable unexpected gzip issue. 
+	// "Request failed. Please analyze the request body for more details. Status code: 401: 
+	// Could not parse JSON: invalid character '\\x1f' looking for beginning of value"
+	req.Header.Set("Accept-Encoding", "identity")
 	req2 := cloneRequest(req) // per RoundTripper contract
 	for _, cookie := range t.SessionObject {
 		req2.AddCookie(cookie)


andygrunwald:
Thanks @hanjm 
Can you point me to the right direction in the docs where this is written down?

hanjm:
@andygrunwald  https://wiki.almworks.com/display/kb/JIRA+with+Apache+Coyote+sending+compressed+reply+without+Content-Encoding+header

andygrunwald:
@hanjm Sadly this seems to be not the official documentation of JIRA / Atlassian. I was not able to find a reference to this in the docs.
I believe that this might be implemented in JIRA itself, but what i know from JIRA that everything is documented quite good.

rbriski:
@hanjm , I think the best way to handle this is just to create your own client with that header set.  I'll use the BasicAuth example because it's a little easier:

```go
// CoyoteTransport is an http.RoundTripper that authenticates all requests
// using HTTP Basic Authentication and sets appropriate headers for Apache Coyote.
type CoyoteTransport struct {
	Username string
	Password string

	// Transport is the underlying HTTP transport to use when making requests.
	// It will default to http.DefaultTransport if nil.
	Transport http.RoundTripper
}

// RoundTrip implements the RoundTripper interface.  We just add the
// basic auth and return the RoundTripper for this transport type.
func (t *CoyoteTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	req2 := cloneRequest(req) // per RoundTripper contract

	req2.SetBasicAuth(t.Username, t.Password)
        req2.Header.Set("Accept-Encoding", "identity") // THIS IS THE CHANGE
	return t.transport().RoundTrip(req2)
}

// Client returns an *http.Client that makes requests that are authenticated
// using HTTP Basic Authentication and sets the correct Apache Coyote headers. 
func (t *CoyoteTransport) Client() *http.Client {
	return &http.Client{Transport: t}
}

func (t *CoyoteTransport) transport() http.RoundTripper {
	if t.Transport != nil {
		return t.Transport
	}
	return http.DefaultTransport
}
```

You should be able to use this with:
```go
tp := CoyoteTransport{
    Username: username,
    Password: password,
}

client, err := jira.NewClient(tp.Client(), jiraURL)
```

WARNING: I haven't tested this code.  It's basically just creating a custom client to pass to the `go-jira` library.

rbriski:
Also, just to cover everything, `401` is an `Unauthorized` code.  Are you sure your credentials are correct for authentication?

PR #133, opened 2018-05-17
Tempo Timesheets woklogs

smgladkovskiy:
Hi!

In my work I need to collect users worklogs. Native Jira methods did not give this information in clean and easy way. But there is a [Tempo Timesheets plugin](https://marketplace.atlassian.com/apps/6572/tempo-timesheets) which has an API [method](http://developer.tempo.io/doc/timesheets/api/rest/latest/#848933329) to collect worklogs by username and dates.

I implemented method to get worklogs. Here it is.


README.md
```diff
@@ -206,7 +206,7 @@ func main() {
 		AuthURL:  fmt.Sprintf("%s/rest/auth/1/session", base),
 	}
 
-	jiraClient, err := jira.NewClient(tp.Client(), base)
+	jiraClient, _ := jira.NewClient(tp.Client(), base)
```
andygrunwald: You dropped `err`, because it is not used in the example, right?
If yes, i would go for an additional 
```go
if err != nil {
	panic(err)
}
```
here, because it is an example and is _really_ often used as a copy / paste go to start code.
With the additional `panic`, people might start think about error handling.

jira.go
```diff
@@ -231,6 +234,10 @@ func (c *Client) Do(req *http.Request, v interface{}) (*Response, error) {
 		// Open a NewDecoder and defer closing the reader only if there is a provided interface to decode to
 		defer httpResp.Body.Close()
 		err = json.NewDecoder(httpResp.Body).Decode(v)
+
+		if err != nil {
+			log.Printf("error JSON encoding body: %v", httpResp.Body)
```
andygrunwald: This leads to a side effect and is not necessary, because a couple of lines after in https://github.com/andygrunwald/go-jira/blob/master/jira.go#L246 the `err` is returned.
I would vote for a removal. Or is there any special reason for it?

jira.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	"github.com/google/go-querystring/query"
 	"github.com/pkg/errors"
+	"log"
```
andygrunwald: See my comment below at `log.Printf...`

worklog.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"time"
+	"strings"
+	"net/url"
+	"fmt"
+)
+
+const TTWorklogTimeFormat = "2006-01-02T15:04:05.000"
+const TTWorklogDateFormat = "2006-01-02"
```
andygrunwald: Can you add comments here? With a link to the docs as well?

worklog.go
```diff
@@ -0,0 +1,111 @@
...
+
+// SprintService handles sprints in JIRA Agile API.
```
andygrunwald: This seems to be a copy / paste doc

README.md
smgladkovskiy: Sure, I will return it

jira.go
smgladkovskiy: Very often I can't understand what kind of error jira thrown when I ask for some data. Logging the answer helped to investigate it. I think that this debugging approach will be helpful for others in interacting with jira.

worklog.go
smgladkovskiy: WorklogBean that returns worklog api uses this time format. Watch the `Returns` section: http://developer.tempo.io/doc/timesheets/api/rest/latest/#848933329

worklog.go
smgladkovskiy: Yep. Sorry. I'll clean it up.

smgladkovskiy:
> And in general, the document strings need to be updated here.
> `godoc` will complain with the current version

@andygrunwald suggest please how to add readme about this functional. Because it is a JIRA plugin api request, I think that it must be some kind of "additional possibilities" section. What do you think?

error.go
```diff
@@ -32,7 +32,7 @@ func NewJiraError(resp *Response, httpError error) error {
 	jerr := Error{HTTPError: httpError}
 	err = json.Unmarshal(body, &jerr)
 	if err != nil {
-		httpError = errors.Wrap(errors.New("Could not parse JSON"), httpError.Error())
+		httpError = errors.Wrap(errors.New(fmt.Sprintf("Could not parse JSON from body %s", string(body))), httpError.Error())
```
rbriski: If you're going to do this just remove `errors.New` and use `fmt.Errorf`

PR #137, opened 2018-05-29
parse timestamps for changelog

Rots:


issue.go
```diff
@@ -53,7 +53,7 @@ type ChangelogItems struct {
 type ChangelogHistory struct {
 	Id      string           `json:"id" structs:"id"`
 	Author  User             `json:"author" structs:"author"`
-	Created string           `json:"created" structs:"created"`
+	Created Time           `json:"created" structs:"created"`
```
rbriski: This will break all code that is expecting a string.  Also, there are no tests to check that this actually works in different situations (bad format, empty string, positive case, etc).

If you want to add a field or possibly just add a function that returns the field as a `Time`.  Something like `CreatedAsTime()`, that would work.

rbriski:
@Rots Any updates here?  Otherwise, I'm going to close this PR.

Rots:
I will add the suggested method to a few places where timestamp is used

Rots:
Not sure how to handle the existing `func (c ChangelogHistory) CreatedTime() (time.Time, error)` notice the return type is `time.Time` instead of `Time`.

ghostsquad:
I agree that adding a helper method here would allow for the desired effect while keeping backwards compatibility.

ghostsquad:
This PR is getting stale. I'll be closing in next week if no one is looking at it.

andygrunwald:
hey @Rots,

thanks for taking the time for this PR.
Any updates on thisß

PR #141, opened 2018-06-07
Support state filter for GetAllSprints call

shawncatz:
I'd like to be able to filter the Sprints received from the GetAllSprints call.

* creates a `GetAllSprintsOptions` structure
* adds a new `GetAllSprintsWithOptions` method to accept an `int` boardID and new options structure
* adds filtering functionality to `GetAllSprintsWithOptions` method
* adds `SprintsList` type for handling pagination results data from `GetAllSprints` request
* updates tests

### DIFF
--- a/board.go
+++ b/board.go
@@ -2,6 +2,7 @@ package jira
 
 import (
 	"fmt"
+	"strconv"
 	"time"
 )
 
@@ -44,9 +45,21 @@ type BoardListOptions struct {
 	SearchOptions
 }
 
-// Wrapper struct for search result
-type sprintsResult struct {
-	Sprints []Sprint `json:"values" structs:"values"`
+// GetAllSprintsOptions specifies the optional parameters to the BoardService.GetList
+type GetAllSprintsOptions struct {
+	// State filters results to sprints in the specified states, comma-separate list
+	State string `url:"state,omitempty"`
+
+	SearchOptions
+}
+
+// SprintsList reflects a list of agile sprints
+type SprintsList struct {
+	MaxResults int      `json:"maxResults" structs:"maxResults"`
+	StartAt    int      `json:"startAt" structs:"startAt"`
+	Total      int      `json:"total" structs:"total"`
+	IsLast     bool     `json:"isLast" structs:"isLast"`
+	Values     []Sprint `json:"values" structs:"values"`
 }
 
 // Sprint represents a sprint on JIRA agile board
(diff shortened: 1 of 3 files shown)


shawncatz:
One other note, when doing a `dep ensure` I noticed there were changes, but I didn't include them in this PR as they weren't related to the changes.

board.go
```diff
@@ -149,9 +157,10 @@ func (s *BoardService) DeleteBoard(boardID int) (*Board, *Response, error) {
 // This only includes sprints that the user has permission to view.
 //
 // JIRA API docs: https://docs.atlassian.com/jira-software/REST/cloud/#agile/1.0/board/{boardId}/sprint
-func (s *BoardService) GetAllSprints(boardID string) ([]Sprint, *Response, error) {
-	apiEndpoint := fmt.Sprintf("rest/agile/1.0/board/%s/sprint", boardID)
-	req, err := s.client.NewRequest("GET", apiEndpoint, nil)
+func (s *BoardService) GetAllSprints(boardID int, options *GetAllSprintsOptions) ([]Sprint, *Response, error) {
```
rbriski: Can you create a different method that is `GetAllSprintsWithOptions` so it doesn't break the interface for older users?

shawncatz: yeah I can do that

shawncatz: I put up a new set of changes. take a look

board.go
```diff
@@ -44,6 +44,23 @@ type BoardListOptions struct {
...
+// GetAllSprintsOptions specifies the optional parameters to the BoardService.GetList
+type GetAllSprintsOptions struct {
+	// State filters results to sprints in the specified states, comma-separate list
+	State string `url:"state,omitempty"`
+
+	SearchOptions
+}
+
+// BoardsList reflects a list of agile boards
+type SprintsList struct {
```
rbriski: Docs don't match struct name.  `SprintsList` vs. `BoardsList`

board.go
```diff
@@ -164,3 +181,24 @@ func (s *BoardService) GetAllSprints(boardID string) ([]Sprint, *Response, error
...
 }
+
+// GetAllSprintsWithOptions will return sprints from a board, for a given board Id and filtering options
+// This only includes sprints that the user has permission to view.
+//
+// JIRA API docs: https://docs.atlassian.com/jira-software/REST/cloud/#agile/1.0/board/{boardId}/sprint
+func (s *BoardService) GetAllSprintsWithOptions(boardID int, options *GetAllSprintsOptions) (*SprintsList, *Response, error) {
+	apiEndpoint := fmt.Sprintf("rest/agile/1.0/board/%d/sprint", boardID)
+	url, err := addOptions(apiEndpoint, options)
+	req, err := s.client.NewRequest("GET", url, nil)
```
rbriski: Also, what do you think about having `GetAllSprints` call this with nil options.  That should avoid some repeated code.

board.go
shawncatz: Apologies, I was rushing before I went into a meeting.

board.go
shawncatz: Yeah, that's a much better idea. I shouldn't be rushing :)

shawncatz:
@rbriski let me know if there's anything else

shawncatz:
this should also resolve #123 (at least in spirit, because it allows to pass pagination into the call)

PR #142, opened 2018-06-08
Add method to retrieve Priorities

Thiht:


andygrunwald:
Thanks for the contribution.
Can you add unit tests for the `GetList` of the `PriorityService`?
Furthermore, the service needs to be added to the main client. See https://github.com/andygrunwald/go-jira/blob/master/jira.go#L37 and https://github.com/andygrunwald/go-jira/blob/master/jira.go#L73

Thiht:
done :)

PR #144, opened 2018-06-25
Add ComponentService to allow creation of Components

natemara:
In my application, I want to programmatically create components. Since this functionality didn't exist in go-jira, I wanted to add it.

I wanted some input on one decision I made - there is already a `Component` type in go-jira used to represent a component that is attached to an issue. Because of this, I created a `FullComponent` type that has all the details of a component added to it already. I thought that was better than creating a backwards-incompatible change by renaming the `Component` type to `IssueComponent` or something like that.

### DIFF
--- a/component.go
+++ b/component.go
@@ -1,0 +1,67 @@
+package jira
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+)
+
+// ComponentService handles components for the JIRA instance / API.
+//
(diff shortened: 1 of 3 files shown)


natemara:
Hold off on this, testing has shown that this may not work exactly correctly.

andygrunwald:
@natemara If you get it working and are confident to push it upstream, feel free to reopen or open a new PR.

PR #146, opened 2018-06-27
Added the method to get the profile of the currently logged in user.

jdevelop:
https://developer.atlassian.com/cloud/jira/platform/rest/#api-api-2-myself-get

Signed-off-by: Eugene Dzhurinsky <jdevelop@gmail.com>


andygrunwald:
@jdevelop Looks good to me.
Would you mind to add a unit test for this method, too?

jdevelop:
@andygrunwald  done :) I was thinking of adding some reflection-based library to check JSON to User marshalling, but perhaps this is the overkill.

user.go
```diff
@@ -99,12 +99,29 @@ func (s *UserService) GetGroups(username string) (*[]UserGroup, *Response, error
...
+	return &user, resp, nil
+}
+
 // Find searches for user info from JIRA:
 // It can find users by email, username or name
 //
 // JIRA API docs: https://docs.atlassian.com/jira/REST/cloud/#api/2/user-findUsers
 func (s *UserService) Find(property string) ([]User, *Response, error) {
-	apiEndpoint := fmt.Sprintf("/rest/api/2/user/search?username=%s", property)
+	apiEndpoint := fmt.Sprintf("/rest/api/2/user/search?username=%s&maxResults=1000", property)
```
rbriski: I don't think we should hardcode config options.  This should be part of a config struct that's passed into every function call.  Look at some of the other services that have config structs for an example.

jdevelop:
Sorry, that was some accident - I am working on getting the better user search with paging, and somehow this artifact has slipped through.

PR #147, opened 2018-06-27
Add MarshalJSON method for Time type

Xjs:
I noticed the Time type has no `MarshalJSON` method when marshaling objects from this library to JSON for storage and later retrieval using `UnmarshalJSON`. `UnmarshalJSON` failed because the marshaled Time fields contained only `{}`, because the Time type had no `MarshalJSON` method.

Having noticed this, I wondered why the unit tests didn't catch this. I noticed that the current design of the unit tests doesn't verify the JSON that is generated in requests with payload (such as `TestIssueService_Create` or `TestIssueService_Update`). The testMux handler in those methods just checks whether method and URL are correct and then responds with a static response. Maybe it would be interesting to add a check whether the request JSON has the correct format as well. (However, I currently lack the time to implement those tests.)

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -344,6 +344,12 @@ func (t *Time) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// MarshalJSON will transform the time.Time into a JIRA time
+// during the creation of a JIRA request
+func (t Time) MarshalJSON() ([]byte, error) {
+	return []byte(time.Time(t).Format("\"2006-01-02T15:04:05.999-0700\"")), nil
+}
+
 // UnmarshalJSON will transform the JIRA date into a time.Time
 // during the transformation of the JIRA JSON response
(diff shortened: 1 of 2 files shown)


Xjs:
Will do, but not this week anymore I guess.

andygrunwald:
@Xjs No hurry. Take your time. We will keep the PR open.

rbriski:
@Xjs I would love if we could get this into `go-jira`.  Can you take a bit of time and write the tests?

rbriski:
:+1:

rbriski:
Thanks for the update.  I've merged it but I'm going to wait to add it to a release.  I think this will actually break old code if people are marshaling the JSON themselves after loading the struct.  I know this is happening because there have been issues opened that specifically ask about it.  I plan on pushing a `v2` release soon because of JIRA dropping cookie authentication.  I'll include this update in it.

PR #148, opened 2018-06-27
Create component service

natemara:
This will allow users to create Jira components with go-jira. I fixed the problems that were in #144, and I actually tested it with Jira this time, so I think we're all good.

### DIFF
--- a/component.go
+++ b/component.go
@@ -1,11 +1,5 @@
 package jira
 
-import (
-	"encoding/json"
-	"fmt"
-	"io/ioutil"
-)
-
 // ComponentService handles components for the JIRA instance / API.
 //
 // JIRA API docs: https://docs.atlassian.com/software/jira/docs/api/REST/7.10.1/#api/2/component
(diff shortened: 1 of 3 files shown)


andygrunwald:
Looks good on the first glance.
We have already the structs `Component` (https://github.com/andygrunwald/go-jira/blob/37f2d5b7594981def803aab0a8f21516afa13f92/issue.go#L255) and `ProjectComponent` (https://github.com/andygrunwald/go-jira/blob/5cfdb85cc91c6299f75b6504a1d0ec174c21be39/project.go#L60).
I wonder how they are different from `FullComponent`?

component.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	resp, err := s.client.Do(req, nil)
+	if err != nil {
+		// incase of error return the resp for further inspection
+		return nil, resp, err
+	}
+
+	component := new(ProjectComponent)
```
rbriski: Is there a special reason that you're doing all of the deserialization yourself? 

`Do` accepts an empty struct so you can pass `component` in.

```go
component := new(ProjectComponent)
resp, err := s.client.Do(req, component)
if err != nil {
  err = NewJiraError(resp, err)
}
return component, resp, err
```

natemara: I was just trying to follow conventions I saw in other places. I figured there was some reason that the serialization was being done this way, for example in `Issue.Create` https://github.com/natemara/go-jira/blob/bd60d4e7dad1573ff3e0886aee5ad82af70c865d/issue.go#L626. If this is not the recommended way to do deserialization, I can fix it.

component.go
```diff
@@ -0,0 +1,34 @@
...
+// Create creates a new JIRA component based on the given options.
+func (s *ComponentService) Create(options *CreateComponentOptions) (*ProjectComponent, *Response, error) {
+	apiEndpoint := "rest/api/2/component"
+	req, err := s.client.NewRequest("POST", apiEndpoint, options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	component := new(ProjectComponent)
+	resp, err := s.client.Do(req, component)
```
rbriski: Can you handle the error here?  `NewJiraError` decorates the error with more information in case downstream libraries want to inspect it

```go
if err != nil {
  err = NewJiraError(resp, err)
}
return component, resp, err
```

PR #149, opened 2018-07-03
Added functions to provide optional parameters to the user search.

jdevelop:
Signed-off-by: Eugene Dzhurinsky <jdevelop@gmail.com>

### DIFF
--- a/user.go
+++ b/user.go
@@ -33,6 +33,15 @@ type UserGroup struct {
 	Name string `json:"name,omitempty" structs:"name,omitempty"`
 }
 
+type userSearchParam struct {
+	name  string
+	value string
+}
+
+type userSearch []userSearchParam
+
+type userSearchF func(userSearch) userSearch
+
 // Get gets user info from JIRA
 //
 // JIRA API docs: https://docs.atlassian.com/jira/REST/cloud/#api/2/user-getUser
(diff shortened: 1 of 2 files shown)


user.go
```diff
@@ -116,12 +125,59 @@ func (s *UserService) GetSelf() (*User, *Response, error) {
...
+	}
+}
+
 // Find searches for user info from JIRA:
 // It can find users by email, username or name
 //
 // JIRA API docs: https://docs.atlassian.com/jira/REST/cloud/#api/2/user-findUsers
-func (s *UserService) Find(property string) ([]User, *Response, error) {
-	apiEndpoint := fmt.Sprintf("/rest/api/2/user/search?username=%s", property)
+func (s *UserService) Find(property string, tweaks ...userSearchF) ([]User, *Response, error) {
```
rbriski: Will changing the function signature affect code that was written before the change?

user.go
```diff
@@ -33,6 +33,15 @@ type UserGroup struct {
 	Name string `json:"name,omitempty" structs:"name,omitempty"`
 }
 
+type userSearchParam struct {
+	name  string
+	value string
+}
+
+type userSearch []userSearchParam
```
rbriski: Not sure if this is worth an entirely new type.  I think the code would be more readable if you just used `[]userSearchParam`.  Renaming it hides the fact that it's a list.

user.go
jdevelop: Nope, as you can see in the original test for this functionality. The API is backwards compatible.

user.go
jdevelop: I started with the `struct` but ended up in just building the query. In this case I think it's a matter of style - I like to name types (coming from Scala world), and this one is unexported. Can change it, no problem - if necessary.

user.go
rbriski: Perfect!

user.go
rbriski: Yeah, I think it's more readable if you just use the `[]userSearchParam` instead of hiding it behind another type.  If someone else uses these functions it'll be more obvious that a list is being returned.

jdevelop: It is only for the **internal** use, and I'd like to restrict the type of function return so people won't be able to pass something very custom with the incompatible unexported type.

just for the sake of purity and type-safety :)

rbriski: I'll merge this but I disagree.  Whether it's use is internal or not, you've created a pretty interesting functionality that, hopefully, someone else will extend.  At  that point, they'll be using your internal types.  

I think the readability of the obvious slice is better than hiding it behind a type that doesn't really add anything but does hide the fact that you're actually returning (or requesting) a slice.

PR #152, opened 2018-07-12
Groups: Implement Create/Delete Groups functions

gappan:
- The Schema for the Group seemed to be different than the one suggested in the [Jira Docs](https://developer.atlassian.com/cloud/jira/platform/rest/#api-api-2-group-get). 
- 2 methods to create /delete a group in Jira

### DIFF
--- a/group.go
+++ b/group.go
@@ -1,7 +1,9 @@
 package jira
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/url"
 )
 
@@ -22,12 +24,11 @@ type groupMembersResult struct {
 }
 
 // Group represents a JIRA group
+// Schema can be found here https://developer.atlassian.com/cloud/jira/platform/rest/#api-api-2-group-get
 type Group struct {
-	ID                   string          `json:"id"`
-	Title                string          `json:"title"`
-	Type                 string          `json:"type"`
-	Properties           groupProperties `json:"properties"`
-	AdditionalProperties bool            `json:"additionalProperties"`
+	Name  string        `json:"name"`
+	Self  string        `json:"self"`
+	Items []GroupMember `json:"items"`
 }
 
 type groupProperties struct {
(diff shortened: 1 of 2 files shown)


group.go
```diff
@@ -152,3 +153,50 @@ func (s *GroupService) Remove(groupname string, username string) (*Response, err
...
+		return nil, nil, err
+	}
+
+	resp, err := s.client.Do(req, nil)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	responseGroup := new(Group)
+	defer resp.Body.Close()
```
rbriski: You don't need to do all of this.  The `Do` function already handles deserializing the JSON into the struct.

group_test.go
```diff
@@ -109,3 +109,45 @@ func TestGroupService_Remove(t *testing.T) {
...
+
+func TestGroupService_Create(t *testing.T) {
+	setup()
+	defer teardown()
+	testMux.HandleFunc("/rest/api/2/group", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
+		testRequestURL(t, r, "/rest/api/2/group")
+
+		w.WriteHeader(http.StatusCreated)
+		fmt.Fprint(w, `{"id": "https://docs.atlassian.com/jira/REST/schema/add-group#","title": "Add Group","type": "object","name" : "hello" ,"additionalProperties": false}`)
```
rbriski: I can't find any documentation like this.  Where is it?  The only schema I see is:
```json
{
  "properties": {
    "name": {
      "type": "string"
    },
    "self": {
      "format": "uri",
      "type": "string"
    },
    "users": {
      "$ref": "#/definitions/PagedListWrapperUserJsonBeanApplicationUser"
    }
  },
  "type": "object"
}
```

rbriski:
@gappan Any more context around this?  Otherwise I'm going to archive this PR and get back to it when I have more time.

PR #159, opened 2018-08-10
Change POST method to PUT method in UpdateComment

Tom-Xie:
Just a quick fix of UpdateComment method


rbriski:
Does it fail when using a `POST`?

Tom-Xie:
I tried to use `POST` in curl, however, the JIRA server did not respond. However after replacing by `PUT`, it worked. Besides, I use this fixed version in my GitHub JIRA synchronization project, and it works correctly.

As I am just a new user of JIRA and this package, I think the issues is that JIRA changed something recently.

rbriski:
Tests are failing.  Can you check them out?

rbriski:
You're right.  I've removed the flaky tests.  As soon as that change goes in, I'll retest this PR and merge.

PR #160, opened 2018-08-10
Updated create issue endpoint in accordance to the documentation

Machiel:
According to the documentation the endpoint is without a trailing slash. With this trailing slash I was unable to create any issues.

Documentation: https://developer.atlassian.com/cloud/jira/platform/rest/#api-api-2-issue-post


andygrunwald:
Thank you.

PR #163, opened 2018-08-23
Add UpdateAssignee method to allow users to change issue assignee

natemara:
This will allow users to update the assignee of an issue.


andygrunwald:
Thank you :)

PR #166, opened 2018-09-05
Support for CookieTransportAuth for IssueService and check for empty cookies

dvdizon:
## Context
_Why does this change need to be made?_

The JIRA instance I am working with prefers the use of cookies for Auth, not `BasicAuth` nor `OAuth`.

As part of my application, I needed a way to have check to see if the current session is expired or not for the `IssueService`. If the session is expired I wanted to know so I can fetch a new session.

Another quirk is for whatever reason, my JIRA instance's session has cookies that have empty values. Because of this, when the `CookieAuthTransport` was getting cookies from the session object, it would carry the empty value cookie into the request and subsequent requests would fail due to invalid cookie sessions.

## Objective

_What is this MR trying to do_

* **Adding the `GetSession()` function to `IssueService`**
By adding the `GetSession()` function, it enables users to use a `CookieAuthTransport` for the `IssueService`. This enables the application to see if it needs to fetch a new session once the session expires.

An example of using this for `IssueService`:

```golang
// Create a new authentication via session cookie
tp := jira.CookieAuthTransport{
	Username: "username",
	Password: "password",
	AuthURL:  "https://my.jira.com/rest/auth/1/session",
}

client, err := jira.NewClient(tp.Client(), "https://my.jira.com")
res, err := client.Issue.GetSession()
// If `res != nil && res.Response != nil && res.Response.StatusCode == 200`, this means it was able to get a valid session
// Otherwise, the session is invalid and need to get a new session (code above)

issue, _, err := client.Issue.Get("FOO-1234")
fmt.Printf("JIRA Ticket Assignee: %v", issue.Fields.Assignee.Name)
```

* **Skip empty value cookies**
As mentioned in the context, this change simply checks if the value of the cookie is empty, and does not add it to the request object.

----------------------------

Hmm..

The build for my pull-request failed with:
```
$ go env
$ go get -t ./...
# golang.org/x/sys/unix
../../../golang.org/x/sys/unix/ioctl.go:18: undefined: runtime.KeepAlive
../../../golang.org/x/sys/unix/ioctl.go:28: undefined: runtime.KeepAlive
The command "go get -t ./..." failed and exited with 2 during .
```

Which is failing for `golang 1.4, 1.5, 1.6` for other pull-requests which doesn't seem to be related to my code change, but with the `go get...` in the beginning of the build.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -543,6 +543,26 @@ func (s *IssueService) Get(issueID string, options *GetQueryOptions) (*Issue, *R
 	return issue, resp, nil
 }
 
+// GetSession gets the session if logged in or not via cookies
+// This enables client.NewRequest to work with CookieAuthTransport for Issues
+// Gets an existing cookie session and returns it
+// The caller should check for error incase the session is invalid
+func (s *IssueService) GetSession() (*Response, error) {
+	apiEndpoint := "rest/auth/1/session"
+	req, err := s.client.NewRequest("GET", apiEndpoint, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := s.client.Do(req, nil)
+	if err != nil {
+		jerr := NewJiraError(resp, err)
+		return resp, jerr
+	}
+
+	return resp, nil
+}
+
 // DownloadAttachment returns a Response of an attachment for a given attachmentID.
 // The attachment is in the Response.Body of the response.
 // This is an io.ReadCloser.
--- a/issue_test.go
+++ b/issue_test.go
@@ -53,6 +53,25 @@ func TestIssueService_Get_WithQuerySuccess(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


dvdizon:
Hmm..

The build failed for my pull-request with:
```
# golang.org/x/sys/unix
../../../golang.org/x/sys/unix/ioctl.go:18: undefined: runtime.KeepAlive
../../../golang.org/x/sys/unix/ioctl.go:28: undefined: runtime.KeepAlive
The command "go get -t ./..." failed and exited with 2 during .
```

Which is failing for `golang 1.4, 1.5, 1.6` for other pull-requests.

dvdizon:
I'm thinking of splitting the two changes I have here. After careful review, `GetSession` for the `IssueService` may need reconsideration and/or revisiting the need for it. I will open a pull-request with just the empty cookie check.

PR #168, opened 2018-09-11
Check for empty cookies for CookieAuthTransport

dvdizon:
## Context
_Why does this change need to be made?_

Splitting up change from #166 

The JIRA instance I am working with prefers the use of cookies for Auth, not `BasicAuth` nor `OAuth`.

For whatever reason, my JIRA instance's session has cookies that have empty values. Because of this, when the `CookieAuthTransport` was getting cookies from the session object, it would carry the empty value cookie into the request and subsequent requests would fail due to invalid cookie sessions.

## Objective

_What is this MR trying to do_

* **Skip empty value cookies**
As mentioned in the context, this change simply checks if the value of the cookie is empty, and does not add it to the request object.

### DIFF
--- a/jira.go
+++ b/jira.go
@@ -375,7 +375,10 @@ func (t *CookieAuthTransport) RoundTrip(req *http.Request) (*http.Response, erro
 
 	req2 := cloneRequest(req) // per RoundTripper contract
 	for _, cookie := range t.SessionObject {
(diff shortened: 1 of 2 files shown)


stjohnjohnson:
@andygrunwald and @rbriski would you be able to give us some feedback on this?

Issue #161, opened 2018-08-10
The ExampleNewClient test is failing on 1.9 and 1.10 intermittently


rbriski:
Travis runs on recent PRs are throwing errors on `ExampleNewClient`

andygrunwald:
It seems to be something related to network or to the external JIRA instance, because if you repeat the tests, sometimes they succeed. So they are flaky somehow.

See the builds here:
- https://travis-ci.org/andygrunwald/go-jira/builds/426393620
- https://travis-ci.org/andygrunwald/go-jira/builds/426151774
- https://travis-ci.org/andygrunwald/go-jira/builds/425847237

PR #169, opened 2018-09-12
Move examples into their own directory

rbriski:
The `example_test.go` didn't use the `go test` tooling at all.  It was mainly to show examples of how code should work.  Since it called out to external resources, it ended up being flaky and failing some of the time.  I've moved these "tests" into the `examples` directory and completely removed the deprecated auth code.


rbriski:
This should fix #161

Issue #174, opened 2018-10-12
Project.Roles doesn't support roles other than "Developers"


Weiss91:
I wanted to get list of all roles in a specific project.
These roles are customized in each jira instance. The existing version of this api does only support the role "Developers".
So i changed the project struct to make the roles field unmarshal customized project roles in a map.
This gives the caller the flexibility to check which roles are present in a project.
I also added a test to check if all given roles in mocks/project.json are unmarshalled in the map.
https://github.com/andygrunwald/go-jira/compare/master...Weiss91:master

Do you think this would be a reasonable representation of the data?
This is my first PR, i would be happy to get some feedback on how to improve. :)

rbriski:
@Weiss91 Can you create a real PR?  It will be easier to work with.  We'll be able to review it and you can make changes in the same branch and they'll show up in the PR.  

Your test is a little confusing.  Why would there be 9 and not some other number of roles (or projects)?

ghostsquad:
Fixed in #176

PR #176, opened 2018-10-15
 Fix of Project.Roles Unmarshal Bug

Weiss91:
Add test case to check for completeness of roles unmarshal.
There should be 9 roles in the test, because of the mock/project.json. There are 9 roles given.
Should i comment this in the test?

### DIFF
--- a/project.go
+++ b/project.go
@@ -36,24 +36,22 @@ type ProjectCategory struct {
 
 // Project represents a JIRA Project.
 type Project struct {
-	Expand       string             `json:"expand,omitempty" structs:"expand,omitempty"`
-	Self         string             `json:"self,omitempty" structs:"self,omitempty"`
-	ID           string             `json:"id,omitempty" structs:"id,omitempty"`
-	Key          string             `json:"key,omitempty" structs:"key,omitempty"`
-	Description  string             `json:"description,omitempty" structs:"description,omitempty"`
-	Lead         User               `json:"lead,omitempty" structs:"lead,omitempty"`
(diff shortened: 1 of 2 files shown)


andygrunwald:
> If we do go ahead with this, I think we'd have to update to v2 since it's a breaking change.

If this leads to a more functional library or removing limitations (like it sounds), I am fine with it.
It just needs to be mentioned in the release notes.

Weiss91:
V2 works for me, can't think of a backwards-compatible way that isn't a terrible hack.

andygrunwald:
@rbriski I would go ahead with this breaking change. Maybe release a new major version here. This can be combined with #181  for example.

One important thing is a migration guide, where we document what are the breaking changes from version to version and how to migrate. The symfony project is quite good at it. See https://github.com/symfony/symfony/blob/master/UPGRADE-5.0.md

ghostsquad:
Project roles are arbitrary, right?
https://confluence.atlassian.com/adminjiracloud/managing-project-roles-776636382.html

Which means, there might not be a "Developer" role in some/most organizations. In fact, I'd guess that this is dead code, as it doesn't actually reflect (or possible to reflect) what roles are actually set up. I'm also ok with merging this breaking change (prior to V2). if these assumptions make sense.

ghostsquad:
Fixes #174

Weiss91:
hey, 
what's the state of this issue? 
Will it be merged before v2? 
If not, what's the planned date for v2?

PR #177, opened 2018-10-15
Adds ability to configure whether updates notify users

spmassot:
Adds a new function `UpdateWithOptions` that allows the user of the library to specify an `UpdateQueryOptions` struct that is passed as query params to the `Edit issue` request

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -686,7 +686,7 @@ func (s *IssueService) UpdateWithOptions(issue *Issue, opts *UpdateOptions) (*Is
 //
 // JIRA API docs: https://docs.atlassian.com/jira/REST/cloud/#api/2/issue-editIssue
 func (s *IssueService) Update(issue *Issue) (*Issue, *Response, error) {
-	apiEndpoint := fmt.Sprintf("rest/api/3/issue/%v", issue.Key)
+	apiEndpoint := fmt.Sprintf("rest/api/2/issue/%v", issue.Key)
 	req, err := s.client.NewRequest("PUT", apiEndpoint, issue)
 	if err != nil {
 		return nil, nil, err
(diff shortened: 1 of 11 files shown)


spmassot:
Hey, thanks for the feedback! I like the latter option, because I feel like these options don't belong in the Issue Struct, since they aren't really part of the Issue, just meta-information about how the transition should occur.

issue.go
```diff
@@ -649,6 +656,32 @@ func (s *IssueService) Create(issue *Issue) (*Issue, *Response, error) {
 	return responseIssue, resp, nil
 }
 
+// UpdateWithOptions updates an issue from a JSON representation,
+// while also specifiying query params. The issue is found by key.
+//
+// JIRA API docs: https://docs.atlassian.com/jira/REST/cloud/#api/2/issue-editIssue
+func (s *IssueService) UpdateWithOptions(issue *Issue, opts *UpdateOptions) (*Issue, *Response, error) {
```
rbriski: Do you mind naming this `UpdateQueryOptions`?  It matches other parts of the code better (https://github.com/andygrunwald/go-jira/blob/master/issue.go#L528).

issue.go
```diff
@@ -649,6 +656,32 @@ func (s *IssueService) Create(issue *Issue) (*Issue, *Response, error) {
...
+	apiEndpoint := fmt.Sprintf("rest/api/3/issue/%v", issue.Key)
+	url, err := addOptions(apiEndpoint, opts)
```
rbriski: Awesome!  Thanks for reusing this.

spmassot:
@rbriski thanks again for the feedback, I think this is much closer to an acceptable implementation. Let me know of other changes I can make to get this merged.

spmassot:
sorry about all these ugly commits - let me know what else I can do to get this merged

issue.go
```diff
@@ -670,6 +682,14 @@ func (s *IssueService) Update(issue *Issue) (*Issue, *Response, error) {
 	return &ret, resp, nil
 }
 
+// Update updates an issue from a JSON representation. The issue is found by key.
+//
+// JIRA API docs: https://docs.atlassian.com/jira/REST/cloud/#api/2/issue-editIssue
+func (s *IssueService) Update(issue *Issue) (*Issue, *Response, error) {
+	opts := &UpdateQueryOptions{}
```
rbriski: You should be able to just send nil instead of an empty `opts` var.

```go
return s.UpdateWithOptions(issue, nil)
```

spmassot:
@rbriski Works!

PR #180, opened 2018-10-30
Manage versions startDate

vsellier:
It seems the start date is missing on the version interface : https://developer.atlassian.com/cloud/jira/platform/rest/v3/#api-group-Version

The goal of this pr is to declare it. It add the field on the tests but I'm not sure if it's enough.


andygrunwald:
LGTM. Thank you :)

PR #181, opened 2018-10-31
add issue field 'Affects Version/s'

Tom-Xie:
I found that `IssueFields` struct has a missing field, which is  `Affects Version/s`

`Affects Version/s` has same struct as `Fix Version/s`, however, with different meaning and JSON key as `versions`

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -101,43 +101,43 @@ type IssueFields struct {
 	//      * "workratio": -1,
 	//      * "lastViewed": null,
 	//      * "environment": null,
-	Expand                        string            `json:"expand,omitempty" structs:"expand,omitempty"`
-	Type                          IssueType         `json:"issuetype,omitempty" structs:"issuetype,omitempty"`
-	Project                       Project           `json:"project,omitempty" structs:"project,omitempty"`
-	Resolution                    *Resolution       `json:"resolution,omitempty" structs:"resolution,omitempty"`
-	Priority                      *Priority         `json:"priority,omitempty" structs:"priority,omitempty"`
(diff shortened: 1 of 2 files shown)


andygrunwald:
Thanks @Tom-Xie.
Can you also extend the unit tests?

Tom-Xie:
Done. 

During the implementation of test, I found both `Affects Version/s` and `Fix Version/s` structs are unnecessary, using `Version` in `version.go` instead.

andygrunwald:
Thank you @Tom-Xie.
For me, this looks good. Also, the change with the `Version` and  `FixVersion` is the better way.

@rbriski How is your opinion here?
Afaik this would be a breaking change because we change types here. I think it is a breaking change in the right direction to avoid duplicate code.

rbriski:
Would this actually break anything?  You should be able to add a struct field but not fill it in, right?

Tom-Xie:
Yes, I think the change is backward compatible, as the extra field could be omitted. @rbriski @andygrunwald

ghostsquad:
Actually, my suggestion doesn't actually make this backward compatible from a client creating a version object to be set in Jira perspective. The only way to do this is to have 2 structs, Version & FixVersion - and accept the unfortunate reality that they have the same fields.

ghostsquad:
Fixed in #209 

@Tom-Xie Thank you for the work you put into this PR! I apologize that I have to close this without merging, I'm trying to ensure that anyone who submits a PR, given it's small enough and meets quality standards can get a merge. But I also want to keep things moving forward, as such, issues/PRs that go stale without response may end up being closed without being merged/addressed.

If you have any additional changes, I'll be attentively watching this repo, and happy to get them merged in!

PR #182, opened 2018-11-07
Add support for context package

asv:
The PR changes the API go-jira adding support for "context" package. This will enable golang applications coordinate cancellation accross many different goroutines.

PR breaks backwards compatibility (added `ctx context.Context` as first parameter to all methods), so maybe it should be part of next major release.

### DIFF
--- a/authentication.go
+++ b/authentication.go
@@ -1,6 +1,7 @@
 package jira
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -56,7 +57,7 @@ type Session struct {
 // JIRA API docs: https://docs.atlassian.com/jira/REST/latest/#auth/1/session
 //
 // Deprecated: Use CookieAuthTransport instead
-func (s *AuthenticationService) AcquireSessionCookie(username, password string) (bool, error) {
+func (s *AuthenticationService) AcquireSessionCookie(ctx context.Context, username, password string) (bool, error) {
 	apiEndpoint := "rest/auth/1/session"
 	body := struct {
 		Username string `json:"username"`
@@ -72,7 +73,7 @@ func (s *AuthenticationService) AcquireSessionCookie(username, password string) 
 	}
 
 	session := new(Session)
-	resp, err := s.client.Do(req, session)
+	resp, err := s.client.Do(ctx, req, session)
 
 	if resp != nil {
 		session.Cookies = resp.Cookies()
@@ -119,7 +120,7 @@ func (s *AuthenticationService) Authenticated() bool {
 //
 // Deprecated: Use CookieAuthTransport to create base client.  Logging out is as simple as not using the
 // client anymore
-func (s *AuthenticationService) Logout() error {
+func (s *AuthenticationService) Logout(ctx context.Context) error {
 	if s.authType != authTypeSession || s.client.session == nil {
 		return fmt.Errorf("no user is authenticated")
 	}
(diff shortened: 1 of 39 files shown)


ghostsquad:
I like this change, but it's definitely major and breaking.

asv:
It's a reason to release a new major version and have a beer! ;)

andygrunwald:
@rbriski What is your opinion about this? I personally likeit. The pattern also follows the https://github.com/google/go-github lib.

rbriski:
I’m pretty conservative when it comes to changes like this. This change will basically make everyone using the library update their code if they want to use the latest version. I’d want a prettt big feature for that kind of requirement and I don’t think context is going to affect most people. My guess is that, if people do change their code, they’ll just use ‘context.Background’ and move on.

Other libraries have added context by including a ‘WithContext’ variant of each method name.  I think that better following the Golang convention of backward compatibility.

ghostsquad:
@rbriski @asv @andygrunwald I started an issue (#200) to talk about V2 changes. I think it makes sense to put this in V2 along with all other breaking changes that anyone might want. Care to comment in that issue (#200) with things you want to see changed?

asv:
@rbriski Thanks for your opinion.  I have prepared the patch does not break backward compatibility: https://github.com/asv/go-jira/tree/context_support_v2. If everything is OK, I can make a force-push in this PR. This will allow us to use the context, as long as you decide how to use it in v2.

ghostsquad:
@asv how do you feel about doing the following:

1. Rename all methods with a suffex `WithContext`
2. Create new methods without the suffix, that calls the WithContext variant and passes in a `TODO` context.

Then in V2, we can basically just drop the suffix and remove the extra methods.

Would that work for you?

Also, when you do this, please ensure you are adhering to the following standards for git commits:

conventionalcommits.org/en/v1.0.0-beta.4/#summary
chris.beams.io/posts/git-commit

I still need to work on a CONTRIBUTING.md file and update the README and all that. But I'm seeing if this makes it easier to produce readable git history and release notes.

suhaibmujahid:
This is a great feature to have. I really need this feature in my application.

@ghostsquad if the changes that you requested are stoping the merge of this PR, I can help in that.

ghostsquad:
@suhaibmujahid yes, this still needs to be done. I think the conversation has happened at this point, and I would love to see the addition of `*Context` methods be introduced.

suhaibmujahid:
@ghostsquad how do you prefer to proceed with this? Should I fork the @asv branch and create a new pull request to include the changes that you requested?

suhaibmujahid:
Also, in this pull request, the use of `context` is implemented by adding the `context` to the request after creating it (`req = req.WithContext(ctx)`) which unnecessary clone of the request and extra memory allocations.

I understand that `google/go-github` is implementing the `context` in this way. But they are doing that to ignore adding the context using a build tag for Google App Engine [see here](https://github.com/google/go-github/blob/4ca0dcc089f42d84097334ba2376686736e28a95/github/with_appengine.go#L18).

I think it is better to simply crate the request with the context (`http.NewRequestWithContext`) and not adding the unnecessary extra overhead by cloning the request after creating it.

ghostsquad:
@suhaibmujahid ya, I think request `RequestWithContext` makes sense.

suhaibmujahid:
@ghostsquad, I implemented the changes that you requested in #276.

andygrunwald:
Hey @asv, have a look at the current master.
https://github.com/andygrunwald/go-jira/pull/276 was merged recently. Thanks for the effort from @suhaibmujahid on this! 🍺 

Thanks, @asv for kicking this off and all the work you put into this!

I will move on and close this PR. The idea of the context package is mentioned in https://github.com/andygrunwald/go-jira/issues/200 (with a link to here).

PR #184, opened 2018-11-13
Executable SearchPages example

writemem:
Thought it might be helpful to others to have an example for the SearchPages method.  Feel free to correct for intended usage etc...


andygrunwald:
Thank you :)

PR #185, opened 2018-11-17
Fix error in the example code in README

is2ei:
I found tiny bug in the example described in README.

```
./main.go:19:9: no new variables on left side of :=
```


andygrunwald:
Thank you.

PR #188, opened 2018-11-20
Add Role service

Weiss91:
Add of RoleService with Get and GetList
Add corresponding tests + mocks

### DIFF
--- a/role_test.go
+++ b/role_test.go
@@ -1,0 +1,56 @@
+package jira
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"testing"
+)
+
+func TestRoleService_GetList(t *testing.T) {
+	setup()
+	defer teardown()
+	testAPIEndpoint := "/rest/api/3/role"
+
+	raw, err := ioutil.ReadFile("./mocks/all_roles.json")
+	if err != nil {
+		t.Error(err.Error())
+	}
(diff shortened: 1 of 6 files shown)


role_test.go
```diff
@@ -0,0 +1,56 @@
+package jira
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"testing"
+)
+
+func TestRoleService_GetList(t *testing.T) {
```
ghostsquad: Please add negative tests (those that validate when an error should be returned)

role_test.go
```diff
@@ -0,0 +1,56 @@
...
+	roles, _, err := testClient.Role.GetList()
+	if roles == nil {
+		t.Error("Expected role list. Role list is nil")
+	}
+	if err != nil {
+		t.Errorf("Error given: %v", err)
+	}
+}
+
+func TestRoleService_Get(t *testing.T) {
```
ghostsquad: Please add negative tests (those that validate when an error should be returned)

ghostsquad:
@Weiss91 I know this is probably a small concern, and let me know if it's not a concern to you before I merge:

Could you rebase/squash you commits while following this standard for your commit message?

https://chris.beams.io/posts/git-commit/

This will allow me to rebase & merge (instead of squash and merge) so that you get full credit for these commits.

PR #194, opened 2018-12-31
Issue 193: only parse application/json content-type as JSON

ebilling:
To address issue around having a good error message when authentication if failing, check the content-type and format the error accordingly.


andygrunwald:
Thank you @ebilling

PR #195, opened 2019-01-14
Fix apiEndpoint creation in UserService.Find

lzakharov:
This PR fixes `UserService.Find` in case of passing `property` with `%` signs (this can happen when `property` string must be URL encoded).


andygrunwald:
Thank you.

PR #199, opened 2019-03-19
Issue#198, Fix panic for unauthorized response with nil error

rekhamitt:
This fixes the panic when a response with invalid status is returned by Jira with nil error.


andygrunwald:
Thank you!

PR #206, opened 2019-04-22
Stopped adding empty params to url

Rambatino:
Currently the query parameters become malformed on usage:

``` go
u, x, err := client.Issue.Search("assignee = currentUser() AND resolution = Unresolved order by updated DESC", &jira.SearchOptions{Expand: "renderedFields"})
```

yields:

``` bash
jql=assignee+%3D+currentUser%28%29+AND+resolution+%3D+Unresolved+order+by+updated+DESC&startAt=0&maxResults=0&expand=renderedFields&fields=&validateQuery=
```

which does not return any data.

This is an example fix that avoids appending those query parameters when they are empty. I couldn't find a neat way of doing that without turning all struct types to pointers. 

It breaks the spec here https://github.com/andygrunwald/go-jira/blob/master/issue_test.go#L607

because it now won't pass empty parameters in. This fixes my use case, but the current implementation is a little gross.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -823,8 +823,22 @@ func (s *IssueService) Search(jql string, options *SearchOptions) ([]Issue, *Res
 	if options == nil {
 		u = fmt.Sprintf("rest/api/2/search?jql=%s", url.QueryEscape(jql))
 	} else {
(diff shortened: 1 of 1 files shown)


ghostsquad:
The build is failing for the PR. can you address the issue?

Rambatino:
Closing in favour of https://github.com/andygrunwald/go-jira/pull/216

PR #214, opened 2019-05-12
Fix board request parameter

### LINKED ISSUE #213: board type list option incombatible with REST API
according to the [docs](https://docs.atlassian.com/jira-software/REST/7.0.4/#agile/1.0/board-getAllBoards) BoardListOptions.BoardType should be mapped to type, now it's boardType.

Because of this, filtering boards based on type does not work.

Shall I create a PR?

xaniasd:
The BoardListOption field BoardType was incorrectly mapped to boardType
instead of type. This commit fixes it. A generic test helper
function (testRequestParams) is added in order to improve the
effectiveness of the unit test.

Fixes #213

### DIFF
--- a/board.go
+++ b/board.go
@@ -35,7 +35,7 @@ type Board struct {
 type BoardListOptions struct {
 	// BoardType filters results to boards of the specified type.
 	// Valid values: scrum, kanban.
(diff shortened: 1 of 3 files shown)


andygrunwald:
Thanks @xaniasd :)

PR #218, opened 2019-06-21
Setting up the NewClient method to accept an interface that gives acc…

benji-vesterby:
This change allows NewClient to accept a mocked http.Client, or a different implementation of http.Client which is what I need it for so that I can interact differently with http.Request methods before they're actually sent to JIRA.

### DIFF
--- a/jira.go
+++ b/jira.go
@@ -15,15 +15,21 @@ import (
 	"github.com/pkg/errors"
 )
 
+// httpClient defines an interface for an http.Client implementation so that alternative
+// http Clients can be passed in for making requests
+type httpClient interface {
+	Do(request *http.Request) (response *http.Response, err error)
+}
+
 // A Client manages communication with the JIRA API.
 type Client struct {
 	// HTTP client used to communicate with the API.
-	client *http.Client
+	client httpClient
 
 	// Base URL for API requests.
 	baseURL *url.URL
 
-	// Session storage if the user authentificate with a Session cookie
+	// Session storage if the user authenticates with a Session cookie
 	session *Session
 
 	// Services used for talking to different parts of the JIRA API.
@@ -52,7 +58,7 @@ type Client struct {
 // As an alternative you can use Session Cookie based authentication provided by this package as well.
 // See https://docs.atlassian.com/jira/REST/latest/#authentication
 // baseURL is the HTTP endpoint of your JIRA instance and should always be specified with a trailing slash.
-func NewClient(httpClient *http.Client, baseURL string) (*Client, error) {
+func NewClient(httpClient httpClient, baseURL string) (*Client, error) {
 	if httpClient == nil {
 		httpClient = http.DefaultClient
 	}


ghostsquad:
@benji-vesterby I love it! Can you please rebase & squash your commits to follow the git commit message standards below, and I'll merge this in and include it in the next release! 

In addition, please write a test that validates backward compatibility.

Thanks

conventionalcommits.org/en/v1.0.0-beta.4/#summary
chris.beams.io/posts/git-commit

andygrunwald:
@benji-vesterby Thanks a lot!

> [...] which is what I need it for so that I can interact differently with http.Request methods before they're actually sent to JIRA.

Can you elaborate on your use-case a little bit more? This will not block this merger. Just curious what you want to do or why this is essential for you (i want to learn :))

benji-vesterby:
@ghostsquad Can you elaborate on what you would like from backwards compatibility testing? The existing unit tests (TestNewClient_WithHttpClient) that use http.Client cover the backwards compatibility right? I don't mind writing the tests I just need to understand what is expected. I can definitely mock a new httpClient and create some tests with a fake httpClient. As for the commit messages, rebase and squashing I'll do that after I've added the tests. 

@andygrunwald So I run a number of integrations as part of my projects and one of them is with JIRA. I create and manage millions of tickets per year (in a single instance) and on average I create ~40k-150k tickets per month. These monthly ticket creations all happen in about 1 hour and I've found that Atlassian has a number of issues surrounding their connection pool when you're exercising their API that quickly. It will initially be OK and run for anywhere from 1-15min before the connection pool spikes regardless of the size of your instance and crashes the JIRA system. In order to circumvent this issue I created a throttle that steps in and allows for my team to build the queries against JIRA as if they can happen immediately but on the way to actually executing the request it throttles the actual number and speed with which they are hitting JIRA. Up to this point I did this in the wrapper library I have surrounding your go-jira driver. Now though I need the same functionality across my other integrations and so I've created a wrapper around http.Client similar to how you do in your JIRA driver and that wrapper now handles the throttling on the Do method so that it can work for any API endpoint without much overhead. Since http.Client automatically implements the httpClient interface that I added to the repo it allows me to substitute in my throttle implementation and limit the requests to the JIRA system and essentially tune the connection to JIRA such that I can execute such massive numbers of requests against JIRA without causing back-end connection pool issues. This is my first contribution to an open source project that is not mine, and will not be my last. I will likely start submitting more pull requests for additional JIRA functionality in the near future, so be on the lookout. ;)

ghostsquad:
@benji-vesterby my mistake. Yes, there's a test that covers the backward compatibility. No need for additional tests.

benji-vesterby:
@ghostsquad I've rebased, squashed and updated my commit message

ghostsquad:
@benji-vesterby Use the imperative mood in your git commit subject:

If applied, this commit will `refactor subsystem X for readability`

benji-vesterby:
@ghostsquad hows that?

ghostsquad:
ugh... I hate GitHub sometimes. Rebase and Merge doesn't keep your verified commit status, unfortunately.

andygrunwald:
@benji-vesterby Thank you for the context :) Seems to make sense

PR #222, opened 2019-07-22
feat: Add AccountType and Locale to User struct

mhupman:


mhupman:
FYI the docs don't have examples for these attributes but they are in the response: https://developer.atlassian.com/cloud/jira/platform/rest/v2/#api-rest-api-2-user-search-get

PR #223, opened 2019-07-23
make NotifyUsers switchable to false

0x434d53:
Since `notifyUsers` is true by default (https://developer.atlassian.com/cloud/jira/platform/rest/v3/?utm_source=%2Fcloud%2Fjira%2Fplatform%2Frest%2F&utm_medium=302#api-rest-api-3-issue-issueIdOrKey-put) it's not possible to set the `notifyUsers` to `false` if the `omitempty` tag is there.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -32,7 +32,7 @@ type IssueService struct {
 
 // UpdateQueryOptions specifies the optional parameters to the Edit issue
 type UpdateQueryOptions struct {
-	NotifyUsers            bool `url:"notifyUsers,omitempty"`
+	NotifyUsers            bool `url:"notifyUsers"`
 	OverrideScreenSecurity bool `url:"overrideScreenSecurity,omitempty"`
 	OverrideEditableFlag   bool `url:"overrideEditableFlag,omitempty"`
 }


ghostsquad:
Here's how you would disable user notifications:

```go
options := &UpdateQueryOptions{
    NotifyUsers: false
}
```

In fact, if you don't omit this field, then by _default_ notifyUsers will be set to `false` (because that's the zero-value of a boolean), instead of the "default" value in Jira (`true`).

ghostsquad:
https://stackoverflow.com/questions/37756236/json-golang-boolean-omitempty

ghostsquad:
I don't think removing `omit:empty` is the right call here, because it will turn off notifications then unless someone specifies this option explicitly.

I think the right course of action is to use a pointer to a boolean, which will then be `nil`, `true` or `false`, with `nil` being the "empty" value, which is then omitted.

ghostsquad:
Although this is a backward-incompatible change, I think this is acceptable, because currently using this option as is, doesn't do anything.

When you update your PR, please follow the following standards for your _single_ commit message.

https://conventionalcommits.org/en/v1.0.0-beta.4/#summary
https://chris.beams.io/posts/git-commit/#seven-rules

an important one often missed is:
https://chris.beams.io/posts/git-commit/#imperative

Please include `BREAKING CHANGE:` in the commit body per the conventionalcommits, just in case anyone is interested to know what happened.

andygrunwald:
hey @0x434d53,

thanks for taking the time for this PR.
Any updates on this?

PR #230, opened 2019-09-03
feat: add support for JWT auth with qsh needed by add-ons

shdunning:
Adding support for auth using JWT's created with QSH fields and signed with a shared secret exchanged during Add-On installation. All other officially supported Atlassian clients have support for this and we found it useful in our environment.

Read more here: https://developer.atlassian.com/cloud/jira/platform/understanding-jwt

Also, added the `go.sum` file per the [best practices](https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file).


rbriski:
Hi @shdunning.  I added support for jira to the core Golang oauth2 library a while ago: https://godoc.org/golang.org/x/oauth2/jira

I thought I had added some docs to this repo but I don't see them.  Maybe I didn't commit them?  Tell me if this takes care of what you need.

rbriski:
Wait, you're a committer on my diff with the oauth2 library so you must already know about it.  I'm guessing this is something different.

ghostsquad:
@rbriski thank you for your contributions! If you could please add tests and rebase/squash commits, while following the below references for your commit messages, I'd be happy to merge this!

https://www.conventionalcommits.org/en/v1.0.0-beta.4/#summary
https://chris.beams.io/posts/git-commit/#seven-rules

shdunning:
@rbriski yeah, this PR is to support server <-> server auth w/out doing the user impersonation, which is what was happening using the oauth2 approach. When you use these JWTs to make API calls, then user actions are registered under a user that is created on behalf of the add-on installation instead of the admin granted the oauth2 token (better for automation).

Also, all of the other client libraries provided by Atlassian have support for this type of auth; namely, constructed the `qsh` field in the claims is code that users shouldn't have to replicate.

shdunning:
ok @ghostsquad this should be ready -- let me know if you need anything else.

ghostsquad:
@shdunning it looks like the build is failing. `strings.ReplaceAll` seems to be unique to `1.12`.

shdunning:
@ghostsquad great catch! i fixed that, but now 1.8 build seems to be failing due to `math/bits` import - not sure if that's related to my change or not.

ghostsquad:
@shdunning ya, I see that. I won't let it block this merge. I did a search for `math/bits` and there doesn't seem to be any direct usage of that in the library. I'm still trying to find time to dig into why/where this is being imported. @andygrunwald if you have time to look at the `math/bits` problem, that would be appreciated.

ghostsquad:
@shdunning very small commit message change request: the sentence "when merged this commit will..." should make sense

currently, `this commit will support for JWT` doesn't read right. If you simply prepend `add` to the commit message (e.g. `feat: add support for...`) that will resolve this issue.

shdunning:
@ghostsquad done.

ghostsquad:
@shdunning please rebase. I decided we should probably just drop support for Go 1.8. This will at least give this build a passing checkmark and make people feel better about it.

andygrunwald:
@ghostsquad Sorry, right now I have very less time to check this issue out (due to some private commitments) :(
I hope you and the community get this managed. Thanks for taking care!

PR #231, opened 2019-09-05
drop go 1.8 support

ghostsquad:


ghostsquad:
@andygrunwald I figured I would add you to this. I think it makes sense to support probably 3 versions of Go, just to help keep things moving forward. `1.13` was just released, so I'm going to add another PR to add that to the build list, but I suspect that it's too early to fully support it.

andygrunwald:
This is fine for me :)
Go is a fast-moving language + backward compatibility.  Thanks for taking care! I owe you a ☕️ or 🍺 once we meet!

Issue #256, opened 2020-02-27
IssueService.GetWatchers does not work


clementlecorre:
Hello!

I have a problem with `GetWatchers`(https://godoc.org/github.com/andygrunwald/go-jira#IssueService.GetWatchers) method.

For some time now, I've been getting 404 errors. After analyzing the code and the problem I think that the key 'name' does not exist anymore in the Jira return payload.

My error message :
```
Status code: 404: Could not parse JSON: unexpected end of JSON input 
```

API query (with curl) : `/rest/api/2/issue/ISSUE-KEY/watchers`
```json
{
  "self": "https://mydomain.atlassian.net/rest/api/2/issue/ISSUE-KEY/watchers",
  "isWatching": false,
  "watchCount": 1,
  "watchers": [
    {
      "self": "https://mydomain.atlassian.net/rest/api/2/user?accountId=000000000000000000000000",
      "accountId": "000000000000000000000000",
      "emailAddress": "user@mydomain.com",
      "avatarUrls": {
        "48x48": "",
        "24x24": "",
        "16x16": "",
        "32x32": ""
      },
      "displayName": "user USER",
      "active": true,
      "timeZone": "Europe/Paris",
      "accountType": "atlassian"
    }
  ]
}
```

IssueService.GetWatchers uses UserService.Get (https://godoc.org/github.com/andygrunwald/go-jira#UserService.Get). Without name key the operation returns something invalid.

Do you have the same problem as me ?

ghostsquad:
hmm, the error doesn't represent a missing field. That error means that response is not valid json.

It's possible you are getting a different response from the library than you are from curl. 404's are sometimes returned instead of a 401, and the body of the response in that case could be HTML.

Can you make other requests successfully from the library?

clementlecorre:
I did my test with curl and the lib go-jira.

`GetWatchers()` consists of two queries.

1. The first one is the rest/api/2/issue/%s/watchers which returns a payload with a missing field (username).
2. The second one I get a 404 error because the username field is missing.

I'm using the cloud jira version.

ghostsquad:
#259 introduces a new method to support the `AccountId`. This might fix your issue?

clementlecorre:
Yes! Thanks :)

PR #259, opened 2020-03-05
fix: UserService.Get change params username to accountId

clementlecorre:
https://github.com/andygrunwald/go-jira/issues/256

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -249,6 +249,7 @@ type Watches struct {
 type Watcher struct {
 	Self        string `json:"self,omitempty" structs:"self,omitempty"`
 	Name        string `json:"name,omitempty" structs:"name,omitempty"`
+	AccountID   string `json:"accountId,omitempty" structs:"accountId,omitempty"`
 	DisplayName string `json:"displayName,omitempty" structs:"displayName,omitempty"`
 	Active      bool   `json:"active,omitempty" structs:"active,omitempty"`
 }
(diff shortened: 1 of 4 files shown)


ghostsquad:
@clementlecorre this looks reasonable, however I'm a bit surprised if this has never worked. Or was this an API change on Jira's end?

ghostsquad:
I see now:
https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-user-privacy-api-migration-guide/

ghostsquad:
@clementlecorre unfortunately, not everyone is going to be running Jira Cloud, or it's possible they haven't updated, and I think this would be a breaking change in that case.

Would you mind instead adding a new method such as `UserService.GetByAccountID` and comment that `Get` will may not work and will be deprecated based on the upstream documentation?

clementlecorre:
Indeed it worked before because I use it regularly.

Indeed it worked before because I use it regularly.
Here is a fix to use the deprecated method if the accountID is not found. The default method is the new (GetByAccountID).

issue.go
```diff
@@ -1236,9 +1237,17 @@ func (s *IssueService) GetWatchers(issueID string) (*[]User, *Response, error) {
 	result := []User{}
 	user := new(User)
 	for _, watcher := range watches.Watchers {
-		user, resp, err = s.client.User.Get(watcher.Name)
-		if err != nil {
-			return nil, resp, NewJiraError(resp, err)
+		if watcher.AccountID != "" {
+			// try fallback deprecated method
```
ghostsquad: I think this comment is in the wrong place, since `GetByAccountID` is not the deprecated method.

clementlecorre: Fixed

ghostsquad:
Can you rebase and squash these commits?

Issue #260, opened 2020-03-05
&maxResults=50 causes no issues to be returned on load-balanced environment


evilnerd:
When using the method ```SearchPages```, the GET query gets suffixed with ```&maxResults=50``` (by default). This works fine on my standalone Jira server, but not on our load-balanced datacenter edition (multi-node) in which case simply no issues are returned even though the query would normally return several issues. 

I tracked the issue down to the ```Search``` method, where the ```&maxResults=50``` is added unescaped. By changing the call to: ```...&amp;maxResults=50``` I could solve the issue (and this keeps working on my single-node instance). 

A second, related issue is that when no Jira Issues are returned by the query, the ```SearchPages``` method ends in a endless loop.

evilnerd:
I have a fix but I'm trying to figure out how to push it here to create pull request.

PR #263, opened 2020-03-13
Fix paging with load balancer going to endless loop

evilnerd:
# PR Description

When used with a load balanced Jira, the SearchPages method would end
up in an endless loop. This was caused by a bug where Jira would not
handle the MaxResults=50 that is sent by defaul properly, thus retur-
ning no issues. The SearchPages method didn't check for empty results
and ended up in an endless loop.

Fixed this by
1. Pre-escaping '&maxResults' to '&amp;MaxResults'.
2. Adding a check in SearchPages to see if the issues array is empty
   before going into the endless 'for'.

Also fixed the appropriate tests.

Fixes issue #260.

# Checklist

* [x] Tests added
  * [x] Good Path
  * [x] Error Path
  * [x] [https://conventionalcommits.org/en/v1.0.0-beta.4/#summary](https://conventionalcommits.org/en/v1.0.0-beta.4/#summary)
  * [x] [https://chris.beams.io/posts/git-commit/#seven-rules](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -965,7 +965,7 @@ func (s *IssueService) Search(jql string, options *SearchOptions) ([]Issue, *Res
 			u += fmt.Sprintf("&startAt=%d", options.StartAt)
 		}
 		if options.MaxResults != 0 {
-			u += fmt.Sprintf("&maxResults=%d", options.MaxResults)
+			u += fmt.Sprintf("&amp;maxResults=%d", options.MaxResults)
 		}
 		if options.Expand != "" {
 			u += fmt.Sprintf("&expand=%s", options.Expand)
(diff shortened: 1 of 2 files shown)


ghostsquad:
@evilnerd your commit message is ALMOST perfect. Here's the only thing you need to change:

old:
```
Fix paging with load balancer
```

new:
```
fix: paging with load balancer
```

The commands to do this are simple though.

```
git commit --amend

git push --force
```

The first command will let you edit your commit message, which will _change_ the commit sha, and require you to force push up to Github.

https://help.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message

evilnerd:
Made the change. Thanks again for your help!

PR #268, opened 2020-04-02
feat(IssueService): allow empty JQL

kfoxb:
# PR Description

This allows the client to make calls with empty JQL specified. My use case is getting all issues in a Jira instance which you typically do by running an empty search. Currently, specifying an empty string makes an invalid URL because the JQL param is always specified. This PR fixes the issue by dynamically building the URL with url.URL and url.Values types according to the options specified.

Note that some tests had to be changed because when you encode url.Values the params get sorted alphabetically.

# Checklist

* [x] Tests added
  * [x] Good Path
  * [x] Error Path
  * [x] [https://conventionalcommits.org/en/v1.0.0-beta.4/#summary](https://conventionalcommits.org/en/v1.0.0-beta.4/#summary)
  * [x] [https://chris.beams.io/posts/git-commit/#seven-rules](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


andygrunwald:
Thanks a lot @kfoxb!

Issue #272, opened 2020-04-14
Field "username" was removed from /user/search by Atlassian


amossc:
When trying to query a user by email address, the code in https://github.com/andygrunwald/go-jira/blob/master/user.go#L212 tries to query by field `username`, but Atlassian apparently has removed that field as part of their GDPR compliance support.

e.g. see the Q&A in https://community.atlassian.com/t5/Jira-questions/The-query-parameter-username-is-not-supported-in-GDPR-strict/qaq-p/1345106 referring to https://confluence.atlassian.com/cloud/blog/2018/06/say-goodbye-to-usernames-in-atlassian-cloud

I'm not 100% sure about the correct resolution, perhaps using `emailAddress` instead, or providing an option to override the query parameter name.

API documentation reference: https://developer.atlassian.com/cloud/jira/platform/rest/v3/?utm_source=%2Fcloud%2Fjira%2Fplatform%2Frest%2F&utm_medium=302#api-rest-api-3-user-search-get

NOTE: I see that `username` is still mentioned in that document, I suspect that this is an omission in the document update.

ghostsquad:
ya, this looks related to some of the changes in #259 

We accept PRs!

cynthiabaran:
Should this issue be closed now?

PR #297, opened 2020-05-19
Removing the use of username field in searching for users

cynthiabaran:
# Description

Searching users by username is now [deprecated](https://developer.atlassian.com/cloud/jira/platform/api-changes-for-user-privacy-announcement/).
This PR is a fix in the method of finding users, which is now using the `query` param. It fixes issue [#272](https://github.com/andygrunwald/go-jira/issues/272)
Users can be searched by email or display name (depending on privacy settings). 

## Example:
The code can still be used in the same way.

```go
users, res, err := client.User.Find("foo@bar.com")
```
or
```go
users, res, err := client.User.Find("foo")
```

# Checklist

  * [x] Good Path
  * [x] Error Path
(all related existing tests were adapted)

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -1365,12 +1365,6 @@ func (s *IssueService) GetWatchersWithContext(ctx context.Context, issueID strin
 			if err != nil {
 				return nil, resp, NewJiraError(resp, err)
 			}
-		} else {
-			// try fallback deprecated method
-			user, resp, err = s.client.User.Get(watcher.Name)
-			if err != nil {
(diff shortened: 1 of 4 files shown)


andygrunwald:
Thank you @cynthiabaran!

penguinlav:
Hi @cynthiabaran! We use self hosted solution and our version has an deprecated api. When we use method `GetWatchersWithContext` it panic, because `watcher.AccountID != ""` always false.
We decided to roll back to the merge version with this issue, everything worked.

PR #298, opened 2020-05-20
for getting all issue types from Atlassian api

praveenpvp:
# Description

## Example:

```go
// Example code

```

# Checklist

  * [ ] Good Path
  * [ ] Error Path
* [ ] I've not made extraneous commits/changes that are unrelated to my change.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -1512,3 +1512,27 @@ func (s *IssueService) AddRemoteLinkWithContext(ctx context.Context, issueID str
 func (s *IssueService) AddRemoteLink(issueID string, remotelink *RemoteLink) (*RemoteLink, *Response, error) {
 	return s.AddRemoteLinkWithContext(context.Background(), issueID, remotelink)
 }
+
+//Get all issues type with context
+func (s *IssueService) GetIssueWithContext(ctx context.Context) ([]IssueType, *Response, error) {
+
+	apiEndpoint := "/rest/api/2/issuetype"
+	req, err := s.client.NewRequestWithContext(ctx, "GET", apiEndpoint, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	IssueList := []IssueType{}
+	resp, err := s.client.Do(req, &IssueList)
+	if err != nil {
+		jerr := NewJiraError(resp, err)
+		return nil, resp, jerr
+	}
+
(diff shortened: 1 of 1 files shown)


issue.go
```diff
@@ -1512,3 +1512,27 @@ func (s *IssueService) AddRemoteLinkWithContext(ctx context.Context, issueID str
 func (s *IssueService) AddRemoteLink(issueID string, remotelink *RemoteLink) (*RemoteLink, *Response, error) {
 	return s.AddRemoteLinkWithContext(context.Background(), issueID, remotelink)
 }
+
+//Get all issues type with context
+func (s *IssueService) GetIssueWithContext(ctx context.Context) ([]IssueType, *Response, error) {
```
andygrunwald, line 1517: This function receives all IssueTypes. Would it make sense to rename the function here?

```suggestion
func (s *IssueService) GetIssueTypesWithContext(ctx context.Context) ([]IssueType, *Response, error) {
```

issue.go
```diff
@@ -1512,3 +1512,27 @@ func (s *IssueService) AddRemoteLinkWithContext(ctx context.Context, issueID str
 func (s *IssueService) AddRemoteLink(issueID string, remotelink *RemoteLink) (*RemoteLink, *Response, error) {
 	return s.AddRemoteLinkWithContext(context.Background(), issueID, remotelink)
 }
+
+//Get all issues type with context
```
andygrunwald, line 1516: Would you mind to add the API docs here as well? Similar to the other go docs?

issue.go
```diff
@@ -1512,3 +1512,27 @@ func (s *IssueService) AddRemoteLinkWithContext(ctx context.Context, issueID str
...
+		jerr := NewJiraError(resp, err)
+		return nil, resp, jerr
+	}
+
+	return IssueList, resp, nil
+}
+
+//Get all issue type
+func (s *IssueService) GetAllIssueType() ([]IssueType, *Response, error) {
+	return s.GetIssueWithContext(context.Background())
```
andygrunwald, line 1537: See function renaming above.

issue.go
praveenpvp: It will return with how many types of issue can be formed in Atlassian

PR #300, opened 2020-05-30
refactor(v2): 🎨  refactor code for v2

ghostsquad:
# Description

This PR introduces "v2" of the codebase, following the GoLang recommended practice of creating a "v2" subdirectory for major version releases.

This is still very much a work in progress, so much so, that **not all the code for v2 is even checked in**, so it won't compile and other problems. Don't worry about that for now.

Notable changes as of 2020-05-30:

- Simplification of the Jira client, by removing/refactoring several of the methods. Part of this is possible from a design decision at this point to support Go 1.14+ (and thus being able to use features specifically from the newer releases)

- More & better error messages. Instead of just returning errors when they happen, they are wrapped to provide more context. This is still a work in progress, as I want to develop "error behaviors", that clients can use to make decisions: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

- Models are separated from "services", in order to make the code more readable. Smaller files means it's easier to find relevant code.

- Started and continuing to work on some code generation using [cheekybits/genny](https://github.com/cheekybits/genny) or for now, a fork of that, that implements a feature to improve usability of the tool [ghostsquad/genny](https://github.com/ghostsquad/genny). For more information, see https://github.com/cheekybits/genny/pull/64

- Also playing around with using [valyala/fastjson](https://github.com/valyala/fastjson) instead of using struct tags. This requires more coding, but also allows for tighter control over how json gets marshalled and unmarshalled, additionally, `fastjson` is **up to 15x faster than the standard encoding/json**. This particular change is driven from the desire to improve the workflow for using customfields, which is a very common use case of Jira.

# Checklist

  * [ ] Good Path
  * [ ] Error Path
* [ ] I've not made extraneous commits/changes that are unrelated to my change.


ghostsquad:
@andygrunwald 

first experimental code generation is working! I purposely chose a simple file, and I'll continue to plug away at this.

This will mean that although I'm using `fastjson` for marshalling & unmarshalling, I can automatically generate the marshalling/unmarshalling code.

At some point, I'll want to support some amount of custom marshalling/unmarshalling logic controlled by struct tags, such as omitEmpty.

Additionally, I'm thinking that we could also automatically generate services too! Which would make it super trivial to ensure services are written the same way, and are bug free.

```
❯ go run ./api/gen-marshalling.go -src ./api/v2/actoruser.go
./api/v2/actoruser.go
ActorUser
---
Field name: AccountID
Ident: string
RENDERED TEMPLATE

package main

import (
	"fmt"
	"github.com/valyala/fastjson"
)

var actorUserJsonParserPool fastjson.ParserPool

// Unmarshal unmarshals JSON string to issue.
func (x *ActorUser) Unmarshal(j string) error {
	p := actorUserJsonParserPool.Get()
	defer actorUserJsonParserPool.Put(p)
	v, err := p.Parse(j)

	if err != nil {
		return fmt.Errorf("failed to parse ActorUser json: %w", err)
	}

	x.AccountID = string(v.GetStringBytes("accountID"))

	return nil
}
```

andygrunwald:
@benjivesterby Any particular reason why this PR was closed?
Sure, it was inactive for quite some time but I do think the proposals are worth looking at a bit more in detail.

This library could benefit from a major improvement.

I am reopening this, to avoid this is going t be forgotten. Ok?

benjivesterby:
@andygrunwald no problem, was just trying to clean out some PRs

Issue #310, opened 2020-08-11
🆘 Please help keep this project alive - Maintainership requested!


ghostsquad:
I think it's safe to say that neither @andygrunwald nor I use this library much, nor have the time to give it proper attention. We've also reached out to Atlassian to see if they would help by either sponsoring or providing test instances of Jira so that we could write better end-2-end tests, but that was also declined.

### If anyone is interested in helping to maintain this library, including working on a v2, please let us know.

I hate seeing new issues roll in, and existing issues go stale. But I just don't have the time to work on this.

fffinkel:
Hi there! I'm very interested in maintaining this project. I'd like to use the library for a work project, and I think it would be a great thing to get involved with. Please let me know how to proceed.

ghostsquad:
@fffinkel - Great! I suppose there's a bit of a "test" that you should pass before getting started here, and @andygrunwald will need to actually give you access. First, please read this:

https://opensource.guide/best-practices/

and let me know what you think of each subtopic (brief brief essay)

additionally, there are a few other things that I think are important for this project:

1. strict backwards compatibility (within a major version)
2. clean git history, and conventional commits
3. documentation!

Are you familiar with these topics, and do you have prior experience in these areas? Can you point us to other projects that you are either the author of, or are maintaining? It's ok if you are new to these topics and haven't practiced them before now. I'm just asking to get a sense of history.

ghostsquad:
additionally, it might be useful to just simply have you get started on working through some existing issues and PRs, etc. And I think that would also give a sense to how you work prior to giving you a "maintainer" role. @andygrunwald what do you think?

I don't want to put too much friction in this process, but also I think it's important to ensure there's a certain level of trust and responsibility to anyone who is a "maintainer".

fffinkel:
I gave the best practices guide a once-over, but I'll give a thorough read tonight.

Regarding the three points:

1. I have experience with this at work, where I've designed and maintained HTTP and code-level APIs. I understand the concepts and (believe I) would know how to handle such a thing if I were to begin working on a v2 for this project.
2. I am fairly experienced with git, having used it for about 10 years. I still blow my foot off from time to time but am able to fix my own with tools like the ref log. I understand concepts like branches vs tags, remotes/upstreams, merging vs rebasing, etc. My personal dotfiles project history is not great, but for projects at work I always follow common best practices for git commits: 50 char capitalized subject line, imperative mood, answering the "why," etc.
[...]

PR #312, opened 2020-08-15
Disable stale bot

fffinkel:
Per discussion in issue #310, the stale bot is now considered a bad user
experience. Remove its configuration to disable it.


fffinkel:
Hm! It appears the integration must be removed as well.

PR #327, opened 2020-11-30
Update travisci go version

fffinkel:
golang.org/x/term no longer supports Go < 1.12 for security reasons. This in turn breaks our build, so let's follow their lead, and include 1.15 while we're touching this.


ghostsquad (approved):

ghostsquad:
@fffinkel probably should update the README.md too. I'd go as far as pushing us to use the latest 2 or 3 versions. Go themselves only support the last 2 major versions. There's little reason why we shouldn't be following their lead.

fffinkel:
@ghostsquad ah, gotcha! I'll do that.

PR #353, opened 2021-02-13
Add options to GetIssuesForSprint

### LINKED ISSUE #352: No query options for GetIssuesForSprint
## Describe the bug
`GetIssuesForSprint` and `GetIssueWithContext` are missing query options.
It prevents lot of useful filtering, like issue assignee, reported, etc..

## To Reproduce
Just try to filter issues by assignee or other filter

## Expected behavior
I would expect `GetIssuesForSprint` get a query options like others functions

## Possible Solution
Add query options to the function

## Your Environment
* go-jira version (git tag or sha): v1.13.0
* Go version (`go version`): 1.15

## Additional context
I'm trying to filter issues by assignee to create a custom board and I find myself needs to do it by code instead if using the filtering Jira is providing by it's API.

oriser:
# Description
Adding query options to `GetIssuesForSprint` and `GetIssueWithContext`

* **The What**: Adding missing query options to `GetIssuesForSprint` and `GetIssueWithContext`
* **The Why**: To be able to use Jira provided filtering by API
* **Type of change**: Missing API query
* **Breaking change**: Yes. Whoever currently uses `GetIssuesForSprint` and will update to this version, will need to pass `nil` (or real query if he wish)
* **Related to an issue**: closes #352 
* **Jira Version + Type**: Jira cloud

## Example:

```go
issues, _, err := j.client.Sprint.GetIssuesForSprint(sprint.ID, &jira.GetIssuesForSprintOptions{
	Jql: fmt.Sprintf(`assignee="%s"`, j.assignee),
})
```

# Checklist

* [X] Unit or Integration tests added
  * [X] Good Path
  * [X] Error Path
* [X] Commits follow conventions described here:
  * [X] [Conventional Commits 1.0.0](https://conventionalcommits.org/en/v1.0.0-beta.4/#summary)
  * [X] [The seven rules of a great Git commit message](https://chris.beams.io/posts/git-commit/#seven-rules)
* [X] Commits are squashed such that
  * [X] There is 1 commit per isolated change
* [X] I've not made extraneous commits/changes that are unrelated to my change.

### DIFF
--- a/sprint.go
+++ b/sprint.go
@@ -23,6 +23,11 @@ type IssuesInSprintResult struct {
 	Issues []Issue `json:"issues"`
 }
 
+type GetIssuesForSprintOptions struct {
+	SearchOptions
+	Jql string `url:"jql,omitempty"`
+}
+
 // MoveIssuesToSprintWithContext moves issues to a sprint, for a given sprint Id.
 // Issues can only be moved to open or active sprints.
 // The maximum number of issues that can be moved in one operation is 50.
@@ -56,10 +61,15 @@ func (s *SprintService) MoveIssuesToSprint(sprintID int, issueIDs []string) (*Re
 // By default, the returned issues are ordered by rank.
 //
 // Jira API Docs: https://docs.atlassian.com/jira-software/REST/cloud/#agile/1.0/sprint-getIssuesForSprint
-func (s *SprintService) GetIssuesForSprintWithContext(ctx context.Context, sprintID int) ([]Issue, *Response, error) {
+func (s *SprintService) GetIssuesForSprintWithContext(ctx context.Context, sprintID int, options *GetIssuesForSprintOptions) ([]Issue, *Response, error) {
 	apiEndpoint := fmt.Sprintf("rest/agile/1.0/sprint/%d/issue", sprintID)
 
-	req, err := s.client.NewRequestWithContext(ctx, "GET", apiEndpoint, nil)
+	url, err := addOptions(apiEndpoint, options)
+	if err != nil {
+		return nil, nil, fmt.Errorf("add options: %w", err)
+	}
+
+	req, err := s.client.NewRequestWithContext(ctx, "GET", url, nil)
 
 	if err != nil {
 		return nil, nil, err
@@ -75,8 +85,8 @@ func (s *SprintService) GetIssuesForSprintWithContext(ctx context.Context, sprin
 }
 
 // GetIssuesForSprint wraps GetIssuesForSprintWithContext using the background context.
-func (s *SprintService) GetIssuesForSprint(sprintID int) ([]Issue, *Response, error) {
-	return s.GetIssuesForSprintWithContext(context.Background(), sprintID)
+func (s *SprintService) GetIssuesForSprint(sprintID int, options *GetIssuesForSprintOptions) ([]Issue, *Response, error) {
+	return s.GetIssuesForSprintWithContext(context.Background(), sprintID, options)
 }
 
 // GetIssueWithContext returns a full representation of the issue for the given issue key.
--- a/sprint_test.go
+++ b/sprint_test.go
@@ -54,7 +54,41 @@ func TestSprintService_GetIssuesForSprint(t *testing.T) {
 		fmt.Fprint(w, string(raw))
 	})
 
-	issues, _, err := testClient.Sprint.GetIssuesForSprint(123)
+	issues, _, err := testClient.Sprint.GetIssuesForSprint(123, nil)
+	if err != nil {
+		t.Errorf("Error given: %v", err)
+	}
+	if issues == nil {
+		t.Error("Expected issues in sprint list. Issues list is nil")
+	}
+	if len(issues) != 1 {
+		t.Errorf("Expect there to be 1 issue in the sprint, found %v", len(issues))
+	}
+
+}
+
+func TestSprintService_GetIssuesForSprint_WithFilter(t *testing.T) {
+	setup()
+	defer teardown()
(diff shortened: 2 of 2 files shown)


andygrunwald:
The CHange is looking good. Thanks @oriser!

I will keep it open for a few days to provide the opportunity for other contributors to review it.
Especially, because this is a breaking change to the existing API (but I think it is a good + manageable one).

andygrunwald (approved):

gueldenstone:
Hi, sorry if I'm wrong. Maybe it's a good idea to make a new function along the lines of "GetIssuesForSprintWithOptions", so it would not be a breaking change and possibly add a "MaxResults" option to the options? Then other options, such as "MaxResults" and "StartAt" could also be implemented?

benjivesterby:
@gueldenstone has a good point. One way this could be done is by something like this

Update the SearchOptions struct to implement the `query.Encoder` interface from the query package using this method `EncodeValues(key string, v *url.Values) error`

Then add a new struct type for the `JQL` string you want added and also implement the `query.Encoder` interface from the query package using the method `EncodeValues(key string, v *url.Values) error`

Then update the API for GetIssuesForSprintWithContext so that it looks like this

```go
func (s *SprintService) GetIssuesForSprintWithContext(ctx context.Context, sprintID int, encoders ...query.Encoder) ([]Issue, *Response, error) {
...
url := apiEndpoint
for _, encoder := range encoders {
 url, err := addOptions(url, encoder)
}
...
```

Since the `encoders` parameter would be variadic it would not break the API for existing users and allow you to extend the functionality. It would also make it flexible to be able to handle additional encoders in the future.

sprint.go
```diff
@@ -23,6 +23,11 @@ type IssuesInSprintResult struct {
 	Issues []Issue `json:"issues"`
 }
 
+type GetIssuesForSprintOptions struct {
```
benjivesterby, line 26: Remove search options embed and implement the query.Encoder interface 

See comment: https://github.com/andygrunwald/go-jira/pull/353#issuecomment-784989306

sprint.go
```diff
@@ -56,10 +61,15 @@ func (s *SprintService) MoveIssuesToSprint(sprintID int, issueIDs []string) (*Re
 // By default, the returned issues are ordered by rank.
 //
 // Jira API Docs: https://docs.atlassian.com/jira-software/REST/cloud/#agile/1.0/sprint-getIssuesForSprint
-func (s *SprintService) GetIssuesForSprintWithContext(ctx context.Context, sprintID int) ([]Issue, *Response, error) {
+func (s *SprintService) GetIssuesForSprintWithContext(ctx context.Context, sprintID int, options *GetIssuesForSprintOptions) ([]Issue, *Response, error) {
```
benjivesterby, line 64: in place of adding options here add `encoders ...query.Encoder`

See comment: https://github.com/andygrunwald/go-jira/pull/353#issuecomment-784989306

sprint.go
```diff
@@ -56,10 +61,15 @@ func (s *SprintService) MoveIssuesToSprint(sprintID int, issueIDs []string) (*Re
...
 	apiEndpoint := fmt.Sprintf("rest/agile/1.0/sprint/%d/issue", sprintID)
 
-	req, err := s.client.NewRequestWithContext(ctx, "GET", apiEndpoint, nil)
+	url, err := addOptions(apiEndpoint, options)
```
benjivesterby, line 67: Update to loop over multiple encoders and use them to update the url 

See comment: https://github.com/andygrunwald/go-jira/pull/353#issuecomment-784989306

sprint.go
```diff
@@ -75,8 +85,8 @@ func (s *SprintService) GetIssuesForSprintWithContext(ctx context.Context, sprin
 }
 
 // GetIssuesForSprint wraps GetIssuesForSprintWithContext using the background context.
-func (s *SprintService) GetIssuesForSprint(sprintID int) ([]Issue, *Response, error) {
-	return s.GetIssuesForSprintWithContext(context.Background(), sprintID)
+func (s *SprintService) GetIssuesForSprint(sprintID int, options *GetIssuesForSprintOptions) ([]Issue, *Response, error) {
```
benjivesterby, line 88: Update with encoders variadic parameter

See comment: https://github.com/andygrunwald/go-jira/pull/353#issuecomment-784989306

sprint.go
```diff
@@ -75,8 +85,8 @@ func (s *SprintService) GetIssuesForSprintWithContext(ctx context.Context, sprin
...
+	return s.GetIssuesForSprintWithContext(context.Background(), sprintID, options)
```
benjivesterby, line 89: this would become `s.GetIssuesForSprintWithContext(context.Background(), sprintID, encoders...)

benjivesterby (changes requested):
This change would also require the implementation of the query.Encoder interface for the SearchOptions struct which is not in the current set of changed files. 

See comment: https://github.com/andygrunwald/go-jira/pull/353#issuecomment-784989306

oriser:
@benjivesterby I can do it NP, but IMHO it better like the original PR, as I saw the rest of the calls are look like this and I think it better to keep the same conventions..
This breaking change it pretty minor as it will only require to add `,nil` to the function call.

benjivesterby (commented):
After reviewing the query library used the Encoder option is not viable. I do not like the fact that this breaks the API as it will likely affect a number of users. I would prefer that the parameter added was variadic in nature and perhaps implemented a type defined in this package for example. 

```
type QueryValues map[string]string
```

This way multiple query values could be passed while keeping the initial implementation if a user doesn't update their function signature. I'm open to suggestions but I feel like breaking the API for this change is unnecessary.

PR #354, opened 2021-02-18
Adding example for Pagination

kp2401075:
# Description
Adding example of pagination to get all issues for given jql.
code was taken from example given by @eparis at https://github.com/andygrunwald/go-jira/issues/55

* **Adding an example that shows how to get all the issues for given JQL with paginaiton
* **This will be helpful for users with usecases where jira's maxResult doen't give all the issues for given JQL
* **Documentation
* **Breaking change**: NO
* **Related to an issue**: https://github.com/andygrunwald/go-jira/issues/55
* **Jira Version + Type**:In-Cloud (Company manged)

## Example:

```go
package main

import (
	"fmt"

	jira "github.com/andygrunwald/go-jira"
)

//GetAllIssues takes a jira client and returns all issues for given JQL
func GetAllIssues(client *jira.Client, searchString string) ([]jira.Issue, error) {
	last := 0
	var issues []jira.Issue = nil
	for {
		opt := &jira.SearchOptions{
			MaxResults: 100,
			StartAt:    last,
		}

		chunk, resp, err := client.Issue.Search(searchString, opt)
		if err != nil {
			return nil, err
		}

		total := resp.Total
		if issues == nil {
			issues = make([]jira.Issue, 0, total)
		}
		issues = append(issues, chunk...)
		last = resp.StartAt + len(chunk)
		if last >= total {
			break
		}
	}
	return issues, nil
}

func main() {
	jiraClient, _ := jira.NewClient(nil, "https://issues.apache.org/jira/")

	// Jira API has limitation as to maxResults it can return at one time.
	// You may have usecase where you need to get all the issues according to jql
	// This is where this example comes in.
	jql := "project = Mesos and type = Bug and Status NOT IN (Resolved)"
	fmt.Printf("Usecase: Running a JQL query '%s'\n", jql)

	issues, err := GetAllIssues(jiraClient, jql)
	if err != nil {
		panic(err)
	}
	fmt.Println(issues)

}

```
@andygrunwald  please help with below section. I'm unsure.
# Checklist

  * [ ] Good Path
  * [ ] Error Path

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,65 @@ func main() {
 	fmt.Printf("Status after transition: %+v\n", issue.Fields.Status.Name)
 }
 ```
+### Get all the issues for JQL with Pagination
+Jira API has limit on maxResults it can return. YOu may have a usecase where you need to get all issues for given JQL.
+This example shows reference implementation of GetAllIssues function which does pagination on Jira API to get all the issues for given JQL
+```go
+package main
 
+import (
+	"fmt"
+
+	jira "github.com/andygrunwald/go-jira"
+)
+
+//GetAllIssues takes a jira client and returns all issues for given JQL
+func GetAllIssues(client *jira.Client, searchString string) ([]jira.Issue, error) {
+	last := 0
+	var issues []jira.Issue = nil
+	for {
+		opt := &jira.SearchOptions{
+			MaxResults: 100,
+			StartAt:    last,
+		}
+
+		chunk, resp, err := client.Issue.Search(searchString, opt)
+		if err != nil {
+			return nil, err
+		}
+
+		total := resp.Total
+		if issues == nil {
+			issues = make([]jira.Issue, 0, total)
+		}
+		issues = append(issues, chunk...)
+		last = resp.StartAt + len(chunk)
+		if last >= total {
+			break
+		}
+	}
+	return issues, nil
+}
+
+func main() {
+	jiraClient, _ := jira.NewClient(nil, "https://issues.apache.org/jira/")
+
+	// Jira API has limitation as to maxResults it can return at one time.
+	// You may have usecase where you need to get all the issues according to jql
+	// This is where this example comes in.
+	jql := "project = Mesos and type = Bug and Status NOT IN (Resolved)"
+	fmt.Printf("Usecase: Running a JQL query '%s'\n", jql)
+
+	issues, err := GetAllIssues(jiraClient, jql)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(issues)
+
+}
+
+
+```
 ### Call a not implemented API endpoint
 
 Not all API endpoints of the Jira API are implemented into *go-jira*.
--- a/README.md
+++ b/README.md
@@ -222,62 +222,10 @@ func main() {
 }
 ```
 ### Get all the issues for JQL with Pagination
-Jira API has limit on maxResults it can return. YOu may have a usecase where you need to get all issues for given JQL.
+Jira API has limit on maxResults it can return. You may have a usecase where you need to get all issues for given JQL.
 This example shows reference implementation of GetAllIssues function which does pagination on Jira API to get all the issues for given JQL
-```go
-package main
 
-import (
-	"fmt"
-
-	jira "github.com/andygrunwald/go-jira"
-)
-
-//GetAllIssues takes a jira client and returns all issues for given JQL
-func GetAllIssues(client *jira.Client, searchString string) ([]jira.Issue, error) {
-	last := 0
-	var issues []jira.Issue = nil
-	for {
-		opt := &jira.SearchOptions{
-			MaxResults: 100,
-			StartAt:    last,
-		}
-
-		chunk, resp, err := client.Issue.Search(searchString, opt)
-		if err != nil {
-			return nil, err
-		}
-
-		total := resp.Total
-		if issues == nil {
-			issues = make([]jira.Issue, 0, total)
-		}
-		issues = append(issues, chunk...)
-		last = resp.StartAt + len(chunk)
-		if last >= total {
-			break
-		}
-	}
-	return issues, nil
-}
-
-func main() {
-	jiraClient, _ := jira.NewClient(nil, "https://issues.apache.org/jira/")
-
-	// Jira API has limitation as to maxResults it can return at one time.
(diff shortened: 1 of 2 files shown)


kp2401075:
@andygrunwald 
Can you please have a look at checklist and advise If I need to take any action.

thanks.

README.md
```diff
@@ -221,7 +221,65 @@ func main() {
 	fmt.Printf("Status after transition: %+v\n", issue.Fields.Status.Name)
 }
 ```
+### Get all the issues for JQL with Pagination
+Jira API has limit on maxResults it can return. YOu may have a usecase where you need to get all issues for given JQL.
```
benjivesterby, line 225: `YOu` should be `You`

README.md
```diff
@@ -221,7 +221,65 @@ func main() {
...
+	jira "github.com/andygrunwald/go-jira"
+)
 
+//GetAllIssues takes a jira client and returns all issues for given JQL
+func GetAllIssues(client *jira.Client, searchString string) ([]jira.Issue, error) {
+	last := 0
+	var issues []jira.Issue = nil
+	for {
+		opt := &jira.SearchOptions{
+			MaxResults: 100,
```
benjivesterby, line 242: MaxResults can go up to 1000, I would suggest specifying that here as a comment

README.md
```diff
@@ -221,7 +221,65 @@ func main() {
...
+import (
+	"fmt"
+
+	jira "github.com/andygrunwald/go-jira"
+)
 
+//GetAllIssues takes a jira client and returns all issues for given JQL
+func GetAllIssues(client *jira.Client, searchString string) ([]jira.Issue, error) {
+	last := 0
+	var issues []jira.Issue = nil
```
benjivesterby, line 239: []jira.Issue will be nil automatically, this should just be `var issues []jira.Issue`

README.md
```diff
@@ -221,7 +221,65 @@ func main() {
...
+			MaxResults: 100,
+			StartAt:    last,
+		}
+
+		chunk, resp, err := client.Issue.Search(searchString, opt)
+		if err != nil {
+			return nil, err
+		}
+
+		total := resp.Total
```
benjivesterby, line 251: this line should move into the `if issues == nil` statement since it's only used once

README.md
```diff
@@ -221,7 +221,65 @@ func main() {
...
+		if issues == nil {
+			issues = make([]jira.Issue, 0, total)
+		}
+		issues = append(issues, chunk...)
+		last = resp.StartAt + len(chunk)
+		if last >= total {
+			break
```
benjivesterby, line 258: I think this would be better as `return issues, nil` instead of `break` since there are other parts of this loop which use return. Either it should all break, or all return IMO.

examples/pagination/main.go
```diff
@@ -0,0 +1,52 @@
+package main
+
+import (
+	"fmt"
+
+	jira "github.com/andygrunwald/go-jira"
+)
+
+//GetAllIssues takes a jira client and returns all issues for given JQL
+func GetAllIssues(client *jira.Client, searchString string) ([]jira.Issue, error) {
```
benjivesterby, line 10: Comments from README should be updated here as well.

examples/pagination/main.go
```diff
@@ -0,0 +1,52 @@
...
+	last := 0
+	var issues []jira.Issue = nil
+	for {
+		opt := &jira.SearchOptions{
+			MaxResults: 100,
```
benjivesterby, line 15: To minimize queries against the API if a user chooses to copy/paste this code this should probably be updated to `1000`

README.md
```diff
@@ -221,7 +221,65 @@ func main() {
...
+		}
+	}
+	return issues, nil
+}
+
+func main() {
+	jiraClient, _ := jira.NewClient(nil, "https://issues.apache.org/jira/")
```
benjivesterby, line 265: the error should not be ignored here since other parts of the example have proper error handling

examples/pagination/main.go
```diff
@@ -0,0 +1,52 @@
...
+		last = resp.StartAt + len(chunk)
+		if last >= total {
+			break
+		}
+	}
+	return issues, nil
+}
+
+func main() {
+	jiraClient, _ := jira.NewClient(nil, "https://issues.apache.org/jira/")
```
benjivesterby, line 38: the error should not be ignored here since other parts of the example have proper error handling

benjivesterby (changes requested):
I think it might be wise to just link to the example in `examples/pagination/main.go` instead of copying into the README. That way there isn't duplication but you have the same effect. Thoughts?

kp2401075:
@benjivesterby Thanks alot for your help, I've made more changes as you requested.

Can you please verify that is satisfy all the requirements.

thanks again for looking into this.

examples/pagination/main.go
kp2401075: Added error handling

examples/pagination/main.go
```diff
@@ -0,0 +1,55 @@
...
+		if issues == nil {
+			issues = make([]jira.Issue, 0, total)
+		}
+		issues = append(issues, chunk...)
+		last = resp.StartAt + len(chunk)
+		if last >= total {
+			return issues, nil
+		}
+	}
+	return issues, nil
```
benjivesterby, line 37: Vet is indicating that this is unreachable so you should be able to remove it.

examples/pagination/main.go
```diff
@@ -0,0 +1,55 @@
...
+// GetAllIssues will implement pagination of api and get all the issues.
+// Jira API has limitation as to maxResults it can return at one time.
+// You may have usecase where you need to get all the issues according to jql
+// This is where this example comes in.
+func GetAllIssues(client *jira.Client, searchString string) ([]jira.Issue, error) {
+	last := 0
+	var issues []jira.Issue
+	for {
+		opt := &jira.SearchOptions{
+			MaxResults: 1000, // Max results can go upto 1000
```
benjivesterby, line 18: `upto` should be `up to`

benjivesterby (commented):
Changes look good except for the vet issue and a typo once those are fixed I'll approve. Also should add a link to the file in the readme as I mentioned in the comment.

kp2401075:
Updated accordingly @benjivesterby Thanks for your help.

benjivesterby (approved):
Looks great!

Issue #350, opened 2021-02-10
Documentation for custom jira fields


pH-T:
Hi!

First: thanks for your work here! Its nice to have a Jira API impl. in Go!

I stumbled upon the `Unknowns` struct field of `IssueFields` and thus I think it should be possible (?) to add custom Jira fields - but how? Can you provide an example?

Thanks,
pH-T

manuelbcd:
Hello @pH-T , do you mean how to retrieve them?  If so, please have a look to issue_test.go (line 776)
https://github.com/andygrunwald/go-jira/blob/master/issue_test.go
here you can check how to retrieve and work with custom fields and just below, with complex custom fields

If you are talking about a different opertation please explain and I will try to help you

Best.

pH-T:
Thanks for your fast reply!

Sorry, I was not precise enough... I was wondering how I would _create_ an issue with custom fields.

Should something like this work?

```
...
    u := tcontainer.NewMarshalMap()
    u["customField1"] = "customValue"

    i := jira.Issue{
        Fields: &jira.IssueFields{
            Description: "Test Issue from Golang",
            Type: jira.IssueType{
                Name: "Task",
            },
            Project: jira.Project{
                Key: "TEST",
            },
            Summary:  "Test-Sum",
            Unknowns: u,
        },
    }

    client.Issue.Create(&i)
    ...
```
I was not able to get this code snipped to run successfully, but this might be because of my local Jira-Testing-Instance...

manuelbcd:
Yes, I think that is the way, let me bring you some examples from other issues

(Source https://github.com/andygrunwald/go-jira/issues/257#issuecomment-595637698 )
```
custome["customfield_10209"] = map[string]string{"value": assessment}

i := jira.Issue{
	Fields: &jira.IssueFields{
		Assignee: &jira.User{
			Name: jira_user,
		},
		Description: utf8toASCII(description),
		Type: jira.IssueType{
			ID: "10300",
		},
		Project: jira.Project{
			ID: "10308",
		},
		Summary: utf8toASCII(summary),
		Unknowns: custome,
	},
}
```

Also have a look to this post and its resolution: https://github.com/andygrunwald/go-jira/issues/117

Pls let us know if that help

pH-T:
Thank you, that helps! So I was on the right track... :)

andygrunwald:
Hey @pH-T and @manuelbcd 
Thanks for stepping in and solving this issue.

[...]

PR #358, opened 2021-03-03
Added example "Create with custom fields"

manuelbcd:
# Description

New example "examples/createwithcustomfields"

* **The What**: New example demonstrating how to create issues with custom fields.
* **The Why**: There were no examples.
* **Type of change**: Improvement/examples
* **Breaking change**: No
* **Related to an issue**: https://github.com/andygrunwald/go-jira/issues/350

## Example:

Simply try examples/createwithcustomfields and provide required data (url, user, password, custom field name, custom field value). 

```bash
go run examples/createwithcustomfields/main.go

```

### DIFF
--- a/examples/createwithcustomfields/main.go
+++ b/examples/createwithcustomfields/main.go
@@ -3,14 +3,13 @@ package main
 import (
 	"bufio"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"strings"
 	"syscall"
 
 	jira "github.com/andygrunwald/go-jira"
 	"github.com/trivago/tgo/tcontainer"
-	"golang.org/x/crypto/ssh/terminal"
+	"golang.org/x/term"
 )
 
 func main() {
@@ -23,7 +22,7 @@ func main() {
 	username, _ := r.ReadString('\n')
 
 	fmt.Print("Jira Password: ")
-	bytePassword, _ := terminal.ReadPassword(int(syscall.Stdin))
+	bytePassword, _ := term.ReadPassword(int(syscall.Stdin))
 	password := string(bytePassword)
 
 	fmt.Print("Custom field name (i.e. customfield_10220): ")
@@ -40,7 +39,7 @@ func main() {
 	client, err := jira.NewClient(tp.Client(), strings.TrimSpace(jiraURL))
 	if err != nil {
 		fmt.Printf("\nerror: %v\n", err)
-		return
+		os.Exit(1)
 	}
 
 	unknowns := tcontainer.NewMarshalMap()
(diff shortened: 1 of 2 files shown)


examples/createwithcustomfields/main.go
```diff
@@ -0,0 +1,76 @@
...
+	r := bufio.NewReader(os.Stdin)
+
+	fmt.Print("Jira URL: ")
+	jiraURL, _ := r.ReadString('\n')
+
+	fmt.Print("Jira Username: ")
+	username, _ := r.ReadString('\n')
+
+	fmt.Print("Jira Password: ")
+	bytePassword, _ := terminal.ReadPassword(int(syscall.Stdin))
```
benjivesterby, line 26: Can you update this to use the new [golang.org/x/term.ReadPassword](https://pkg.go.dev/golang.org/x/term#ReadPassword) which is part of `golang.org/x/term` instead since `golang.org/x/crypto/ssh/terminal` was deprecated? This new implementation seems to add support for windows as well which the previous version did not.

examples/createwithcustomfields/main.go
```diff
@@ -0,0 +1,76 @@
...
+
+	tp := jira.BasicAuthTransport{
+		Username: strings.TrimSpace(username),
+		Password: strings.TrimSpace(password),
+	}
+
+	client, err := jira.NewClient(tp.Client(), strings.TrimSpace(jiraURL))
+	if err != nil {
+		fmt.Printf("\nerror: %v\n", err)
+		return
```
benjivesterby, line 43: This should be `os.Exit(1)` instead of return to indicate error

examples/createwithcustomfields/main.go
```diff
@@ -0,0 +1,76 @@
...
+				Key: "PROJ1",
+			},
+			Summary:  "Just a demo issue",
+			Unknowns: unknowns,
+		},
+	}
+
+	issue, resp, err := client.Issue.Create(&i)
+	if err != nil {
+		_, err := ioutil.ReadAll(resp.Body)
```
benjivesterby, line 71: The error which occurs on line 69 is being hidden by the error assignment on line 71. This should be updated so that both errors are printed.

manuelbcd:
Thank you for this nice feedback @benjivesterby. I will update the PR and take the opportunity to submit new PRs applying these changes to the other existing examples

manuelbcd:
@benjivesterby FYI I have fixed your two first comments. For the third one "_, err := ioutil.ReadAll(resp.Body)" I have restored the error managment to be similar to "create" example, I can evolve it in another PR if you want.

benjivesterby (approved):
This looks great @manuelbcd thank you for all the hard work!

PR #363, opened 2021-03-16
pass context parameters through

asuffield:
Wrapper functions that take a context should call the version which takes a context, rather than silently switching to context.Background()


benjivesterby:
@asuffield I think this should probably check if the incoming ctx is nil and replace it with the background ctx in that case though right?

asuffield:
> @asuffield I think this should probably check if the incoming ctx is nil and replace it with the background ctx in that case though right?

That's not normal, or done elsewhere here. The usual rule is just "Do not pass a nil context" (https://golang.org/pkg/context/).

Issue #368, opened 2021-03-23
Add WarningMessages to query response to be able to manage warnings


manuelbcd:
## Is your feature request related to a problem? Please describe.
No, it is not a problem.

## Describe the solution you'd like
When using Issue.Search(...) I have noticed that the optional field "WarningMessages" is not implemented in the response struct
struct:
![image](https://user-images.githubusercontent.com/10448961/112126909-32153180-8bc5-11eb-92b2-01e79c313b76.png)
real response:
![image](https://user-images.githubusercontent.com/10448961/112126938-3a6d6c80-8bc5-11eb-97f3-b6c10de7a3f5.png)

The drawback is that you are unable to detect these warnings since most of times the HTTP response is 200 even if one of the fields from the query is a wrong value (i.e. user not exists)

## Describe alternatives you've considered
I think there are no alternatives.

## Additional context
No additional context

@andygrunwald @benjivesterby   what do you think? do you want me to implement it?

andygrunwald:
Hey @manuelbcd,
thanks!

Two questions:
1. Are you talking about the OnPremise JIRA Version (+ which version in particular?) or the Cloud-Version?
2. Do you have a link to the response documentation?

Would love to review a PR for this!

manuelbcd:
I have tested it on Jira Cloud but I think it is part of response (only when something goes wrong) in V3 API on-prem also

You can check the response example in issue-search documentation:

https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/

![image](https://user-images.githubusercontent.com/10448961/112128881-24f94200-8bc7-11eb-8c7f-f4470fb7968a.png)

andygrunwald:
Great, thanks. Happy to review a PR for this.

manuelbcd:
I'm worried my PR is breaking compatibility since I'm adding a warningMessages[] field to issue.Search() response.
Please suggest if there could be a better approach to implement it or if you simply consider it is not worth to add warningMessages considering drawbacks.

andygrunwald:
Thanks, @manuelbcd! I will have a look. Let us continue the discussion about the implementation approach to https://github.com/andygrunwald/go-jira/pull/369

andygrunwald:
Hey,

I am very sorry that this issue has been open for a long time with no final solution. We work on this project in our spare time, and sometimes, other priorities take over. This is the typical open source dilemma.

#### However, there is news: **We are kicking off v2 of this library 🚀** 

[...]

PR #369, opened 2021-03-23
#368 Added warningMessages to issue.search() response

manuelbcd:
# Description

When using Issue.Search(...) response can contain optional field "WarningMessages" that was not implemented in the response struct.

This feature allows to manage possible WarningMessages since jql errors are not rendered as errors but as warning messages.
Without this option, you are unable to find out if your jql query failed because i.e. specified user does not exists.
More information in #368 

* **The What**: Adding WarningMessages[] to searchResult (search response wrapper)
* **The Why**: It is described at #368 (i.e. user specified in jql query does not exist)
* **Type of change**: New feature.
* **Breaking change**: Yes. My implementation adds a new field to issue.Search() return. **Please suggest me if you think there is a better implementation approach.***
* **Related to an issue**: #368

## Example:

```go
chunk, resp, err, warningMessages := client.Issue.Search(searchString, opt)
if err != nil {
	return nil, err
}
if len(warningMessages) > 0 {
  fmt.Printf("Warning messages in response:\n")
  for i, warn := range warningMessages {
	fmt.Printf("Warning %d: %s\n", i, warn)
  }
}

```

# Checklist

  * [ ] Good Path
  * [ ] Error Path


kusuriya:
LGTM, but we may want to have some notification around merging it. It looks like it changes the behavior around search as defined in the code that uses the library and could possibly break peoples code unless they handle for the new output variable. thoughts @andygrunwald

manuelbcd:
Yes @kusuriya, i do agree. Im worried because im not sure if is worth adding this feature while breaking code compatibility. Actually I was asking about that in #368 

If someone can give it a thought... maybe there could be a different approach

andygrunwald:
Thanks for the effort, the PR, and for the enhancement.
Indeed, the current implementation is breaking the compatibility.
We have a few PRs open that will break the existing interfaces, and this is the main reason why those doesn't have been merged yet.
For quite some time, we are talking about an official v2, that fixes many of these implementation details. I see two options here:
1. Keeping this PR for v2 (no timeline yet)
2. Changing the implementation into a compatible way

About 2), one idea would be to create a new Search function that supports these warning messages.
When a new Search functionality is created, I would even suggest changing the type that is returned to something like 
```go
type SearchResult struct {
    Issues []issue ...
    WarningMessage []messages ...
    ....
}
```

By this, we would not break the compatibility in the future, if a new field is part of the structure.
I am curious about what you think and your feedback.

manuelbcd:
Thanks for asking.  In my opinion, adding a new function just to bring the ability to access to hypothetical warning messages could be a bit complex for library consumers...   I bet for the 1st one, implement it into v2 refining implementation details first.  
However if we receive more issues and votes to add the feature to v1, we can always add this third function in the future (in 3 weeks, in 3 months, it does not  matter).

andygrunwald:
Hey,

I am very sorry that this pull request has been open for a long time with no final feedback or merge/decline. We work on this project in our spare time, and sometimes, other priorities take over. This is the typical open source dilemma.

#### However, there is news: **We are kicking off v2 of this library 🚀** 

To provide visibility, we created the [Road to v2](https://github.com/andygrunwald/go-jira/milestone/1) Milestone and calling for your feedback in https://github.com/andygrunwald/go-jira/issues/489

The development will take some time; however, I hope you can benefit from the changes.
If you seek priority development for your pull request + you like to sponsor it, please contact me.

#### What does this mean for my pull request?

We will work on this pull request indirectly.
We might merge it during the development or pull parts of it into the new version.
This means that during the development phase, we aim to tackle it.
Maybe in a different way like it is currently handled.
Please understand that this will take a while because we are running this in our spare time.

#### Final words

Thanks for using and contributing to this library.
If there is anything else you would like to tell us, let us know!

Issue #373, opened 2021-03-30
assignee field while creating jira ticket fails with 400


ptripathi-twilio:
## Describe the bug

When I try to assign a ticket to some while creating it, zapi fails with 4000 bad request.

 jira create API payload log looks like this:
`"fields":{"assignee":{"Password":"","name":"svc.koalat-jira"},"issuetype":{"name":"Test"},"labels":["MyMy"],"project":{"id":"23432"},"reporter":{"Password":"","name":"fgvg.rtf"},"summary":"TestVoid"}}`

but I get a 400 error with msg like this:

`{"errorMessages":[],"errors":{"assignee":"Field 'assignee' cannot be set. It is not on the appropriate screen, or unknown."} `

## To Reproduce

create a ticket with assignee field as `assignee`

## Expected behavior

should create the ticket

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* go-jira version (git tag or she): latest
* Go version (`go version`): latest
* Jira type (cloud or on-premise): on prem

Dynatrace-Adam-Gardner:
Assignee and Reporter need to be user IDs not usernames

ctreminiom:
It seems the user you're using to authenticate the call doesn't have the **Assignable User** permission or the fields are not in the **Issue Create** screen.  @Dynatrace-Adam-Gardner the User ID (account-ID) is required only on the Cloud version, on Jira On-Prem is the _Username_ value

manuelbcd:
@benjivesterby, @andygrunwald  I think this issue is resolved and can be closed

Best.

benjivesterby:
@manuelbcd Sounds good

Issue #69, opened 2017-04-17
Problem with example create


ypsman:
Im trying to Connect with the Example to Jira and creating an Issue.
Getting this as Response 
 {"errorMessages":[],"errors":{"project":"project is required"}}
panic: runtime error: invalid memory address or nil pointer dereference

go version go1.6.2 linux/amd64
Jira 7.3 & 7.1

Im new to golang so a little hint what to do would be nice :)

andygrunwald:
Hey @ypsman,
can you please post your complete code here? Without the source code it is hard to follow. Thanks.

ypsman:
```package main

import (
	"fmt"

	"github.com/andygrunwald/go-jira"
)

func main() {
	jiraClient, err := jira.NewClient(nil, "https://myjira.de/")
	if err != nil {
		panic(err)
	}

	res, err := jiraClient.Authentication.AcquireSessionCookie("APIuser", "pass")
	if err != nil || res == false {
		fmt.Printf("Result: %v\n", res)
		panic(err)
	}

	i := jira.Issue{
		Fields: &jira.IssueFields{
			Assignee: &jira.User{
				Name: "myname",
			},
			Reporter: &jira.User{
				Name: "APIuser",
			},
			Description: "Test Issue",
			Type: jira.IssueType{
				ID: "60",
			},
			Project: jira.Project{
				Name: "APITEST",
			},
			Summary: "Just a demo issue",
		},
	}

	issue, _, err := jiraClient.Issue.Create(&i)

	if err != nil {
		panic(err)
	}

	fmt.Printf("%s: %+v\n", issue.Key, issue.Fields.Summary)
}
```
its just like the Example .

andygrunwald:
Does the Project `APITEST` exists?

ypsman:
Yes, the Fields are correct, its work with a Pyton script.

andygrunwald:
Can you may post the Python script as well? With this i might be able to find the difference and maybe we hit a bug.

ypsman:
Its just a simple script with the Jira Module(python3)
I cant post it complete, but heres the class i use for.
```
from jira import JIRA
from jira.exceptions import JIRAError
class JiraAPI:
    def __init__(self):
        try:
            self.api = JIRA(jiraserver, basic_auth=(jiraauth))
        except JIRAError as e:
            print("Exit Code: " + e.status_code)
            sys.exit()
def ticket_create(self):

    self.ticket = {
                    'project' : self.Project,
[...]

PR #374, opened 2021-04-17
#69 #373 example/create changed to use accountID intead of username

manuelbcd:
Changed creation example to refer reporter by accountID instead of username since it is deprecated in Jira cloud.
Working by accountID is compatible with both Jira Cloud and Jira Server versions.

# Description

* **The What**: examples/create   Changed username field by accountID
* **The Why**: Jira Cloud API is rejecting usernames  from 2018 due to GDPR compliance 
* **Type of change**: Minor fix in an example
* **Breaking change**: Non breaking
* **Related to an issue**: #69 and #373
* **Jira Version + Type**: Jira Cloud

## Example:

Just try examples/create  using go run

### DIFF
--- a/examples/create/main.go
+++ b/examples/create/main.go
@@ -38,7 +38,7 @@ func main() {
 	i := jira.Issue{
 		Fields: &jira.IssueFields{
 			Assignee: &jira.User{
-				Name: "myuser",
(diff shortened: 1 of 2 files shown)


benjivesterby (approved):
Looks great! Thank you for this update.

PR #391, opened 2021-07-30
feat: add customer and request endpoints for service desk api

iamjem:
* **The What**: This adds endpoints for customers and requests in the service desk api.
* **The Why**: Because these endpoints are missing (and I need them in my project).
* **Type of change**: New feature.
* **Breaking change**: This change should be backwards compatible.
* **Related to an issue**: Not that I'm aware of.

## Example:

Tests exist for all new endpoints, which is probably the best source of sample code.

# Checklist

  * [x] Good Path
  * [ ] Error Path
  * [x] [Conventional Commits 1.0.0](https://conventionalcommits.org/en/v1.0.0-beta.4/#summary)

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -1,0 +1,77 @@
+package jira
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+// CustomerService handles ServiceDesk customers for the Jira instance / API.
+type CustomerService struct {
+	client *Client
+}
+
+// Customer represents a ServiceDesk customer.
+type Customer struct {
+	AccountID    string    `json:"accountId,omitempty" structs:"accountId,omitempty"`
+	Name         string    `json:"name,omitempty" structs:"name,omitempty"`
+	Key          string    `json:"key,omitempty" structs:"key,omitempty"`
+	EmailAddress string    `json:"emailAddress,omitempty" structs:"emailAddress,omitempty"`
+	DisplayName  string    `json:"displayName,omitempty" structs:"displayName,omitempty"`
+	Active       *bool     `json:"active,omitempty" structs:"active,omitempty"`
+	TimeZone     string    `json:"timeZone,omitempty" structs:"timeZone,omitempty"`
+	Links        *SelfLink `json:"_links,omitempty" structs:"_links,omitempty"`
+}
+
+// CustomerListOptions is the query options for listing customers.
+type CustomerListOptions struct {
+	Query string `url:"query,omitempty"`
+	Start int    `url:"start,omitempty"`
+	Limit int    `url:"limit,omitempty"`
+}
+
+// CustomerList is a page of customers.
+type CustomerList struct {
+	Values  []Customer `json:"values,omitempty" structs:"values,omitempty"`
+	Start   int        `json:"start,omitempty" structs:"start,omitempty"`
+	Limit   int        `json:"limit,omitempty" structs:"limit,omitempty"`
+	IsLast  bool       `json:"isLastPage,omitempty" structs:"isLastPage,omitempty"`
+	Expands []string   `json:"_expands,omitempty" structs:"_expands,omitempty"`
+}
+
+// CreateWithContext creates a ServiceDesk customer.
+//
+// https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-customer/#api-rest-servicedeskapi-customer-post
+func (c *CustomerService) CreateWithContext(ctx context.Context, email, displayName string) (*Customer, *Response, error) {
+	apiEndpoint := "rest/servicedeskapi/customer"
+
+	payload := struct {
+		Email       string `json:"email"`
+		DisplayName string `json:"displayName"`
+	}{
+		Email:       email,
+		DisplayName: displayName,
+	}
+
+	req, err := c.client.NewRequestWithContext(ctx, "POST", apiEndpoint, payload)
+	if err != nil {
+		return nil, nil, err
+	}
+	resp, err := c.client.Do(req, nil)
+	if err != nil {
+		return nil, resp, NewJiraError(resp, err)
+	}
+	defer resp.Body.Close()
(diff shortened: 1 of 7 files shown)


customer.go
```diff
@@ -0,0 +1,77 @@
...
+	Limit   int        `json:"limit,omitempty" structs:"limit,omitempty"`
+	IsLast  bool       `json:"isLastPage,omitempty" structs:"isLastPage,omitempty"`
+	Expands []string   `json:"_expands,omitempty" structs:"_expands,omitempty"`
+}
+
+// CreateWithContext creates a ServiceDesk customer.
+//
+// https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-customer/#api-rest-servicedeskapi-customer-post
+func (c *CustomerService) CreateWithContext(ctx context.Context, email, displayName string) (*Customer, *Response, error) {
+	apiEndpoint := "rest/servicedeskapi/customer"
```
frzifus, line 46:
```suggestion
	const apiEndpoint = "rest/servicedeskapi/customer"
```

customer.go
```diff
@@ -0,0 +1,77 @@
...
+
+	payload := struct {
+		Email       string `json:"email"`
+		DisplayName string `json:"displayName"`
+	}{
+		Email:       email,
+		DisplayName: displayName,
+	}
+
+	req, err := c.client.NewRequestWithContext(ctx, "POST", apiEndpoint, payload)
```
frzifus, line 56:
```suggestion
	req, err := c.client.NewRequestWithContext(ctx, http.MethodPost, apiEndpoint, payload)
```

customer.go
```diff
@@ -0,0 +1,77 @@
...
+	responseCustomer := new(Customer)
+	if err := json.NewDecoder(resp.Body).Decode(responseCustomer); err != nil {
+		return nil, resp, fmt.Errorf("could not unmarshall the data into struct")
+	}
+
+	return responseCustomer, resp, nil
+}
+
+// Create wraps CreateWithContext using the background context.
+func (c *CustomerService) Create(email, displayName string) (*Customer, *Response, error) {
```
frzifus, line 75: Is there a reason why we expose this method without a context?

request.go
```diff
@@ -0,0 +1,136 @@
...
+// https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-request/#api-rest-servicedeskapi-request-issueidorkey-comment-post
+func (r *RequestService) CreateCommentWithContext(ctx context.Context, issueIDOrKey string, comment *RequestComment) (*RequestComment, *Response, error) {
+	apiEndpoint := fmt.Sprintf("rest/servicedeskapi/request/%v/comment", issueIDOrKey)
+
+	req, err := r.client.NewRequestWithContext(ctx, "POST", apiEndpoint, comment)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	resp, err := r.client.Do(req, nil)
```
james-cast, line 118:
```suggestion
        responseComment := new(RequestComment)
	resp, err := r.client.Do(req, responseComment)
```

request.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return nil, resp, NewJiraError(resp, err)
+	}
+
+	defer resp.Body.Close()
```
james-cast, line 123: can remove lines 123-128 if you use `client.Do` to unmarshal the response.

servicedesk.go
```diff
@@ -112,3 +117,107 @@ func (s *ServiceDeskService) RemoveOrganizationWithContext(ctx context.Context,
...
+			return nil, nil, err
+		}
+		req.URL.RawQuery = q.Encode()
+	}
+
+	resp, err := s.client.Do(req, nil)
+	if err != nil {
+		return nil, resp, NewJiraError(resp, err)
+	}
+	defer resp.Body.Close()
```
james-cast, line 210: Same here, duplicating work of `client.Do`

customer.go
iamjem: There's a `CustomerService.CreateWithContext` which is the context version of the method. I followed this pattern from other service endpoints in the package.

iamjem:
Thank you for the review! I believe that I've addressed your comments, please let me know if there's any other changes needed.

customer.go
frzifus: I think there was more then that :)

servicedesk.go
```diff
@@ -112,3 +117,107 @@ func (s *ServiceDeskService) RemoveOrganizationWithContext(ctx context.Context,
 func (s *ServiceDeskService) RemoveOrganization(serviceDeskID int, organizationID int) (*Response, error) {
 	return s.RemoveOrganizationWithContext(context.Background(), serviceDeskID, organizationID)
 }
+
+// AddCustomersWithContext adds customers to the given service desk.
+//
+// https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-servicedesk/#api-rest-servicedeskapi-servicedesk-servicedeskid-customer-post
+func (s *ServiceDeskService) AddCustomersWithContext(ctx context.Context, serviceDeskID int, acountIDs ...string) (*Response, error) {
```
james-cast, line 124: I created a PR to change `serviceDeskID int` to `serviceDeskID interface{}` because it should have actually been a string, but for the sake of not breaking other peoples use I opt'd for an interface. I would suggest also accepting string values.

See the PR here https://github.com/andygrunwald/go-jira/pull/419

benjivesterby: I would agree with @james-cast, Please update as requested

benjivesterby (changes requested):
Please update as requested by @james-cast since the service desk id is supposed to be a string.

iamjem:
I have updated `serviceDeskID int` to `serviceDeskID interface{}` for the remaining customer methods, and added test coverage.

benjivesterby (approved):
LGTM

Issue #401, opened 2021-09-23
Update to the latest version (v4) "github.com/dgrijalva/jwt-go" 


sitikanthasezzle:
The dependency "github.com/dgrijalva/jwt-go" has a vulnerability, CVE-2020-26160. It would be good to update to the latest version (v4) .

https://github.com/dgrijalva/jwt-go/releases/tag/v4.0.0-preview1

sitikanthasezzle:
@benjivesterby thank you so much

Issue #400, opened 2021-09-23
Update jwt-go library from dgrijalva/jwt-go to form3tech-oss/jwt-go


sitikanthasezzle:
https://github.com/form3tech-oss/jwt-go/releases/tag/v3.2.1
https://nvd.nist.gov/vuln/detail/CVE-2020-26160#vulnCurrentDescriptionTitle
https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMDGRIJALVAJWTGO-596515

PR #382, opened 2021-06-22
update jwt-go to v4

TheAxZim:
# Description

Updating the jwt-go library. Current version is vulnerable: https://www.whitesourcesoftware.com/vulnerability-database/CVE-2020-26160  -- this is a preview version but works the same way. 

Information that is useful here:
* **The What**: Updating a library which is vulnerable.
* **The Why**: Security
* **Type of change**: Dependency update
* **Breaking change**: No
* **Related to an issue**: #343 
* **Jira Version + Type**: Cloud

Issue #343, opened 2021-02-08
Consider upgrading version of jwt-go to fix security vulnerability


lodi-g:
Hi,

The dependency jwt-go has a security vulnerability in all versions <4.0.0: https://github.com/dgrijalva/jwt-go/issues/428 .
Please consider upgrading to the latest version (v4.0.0-preview1) or switching for another library.

Thanks!

benjivesterby:
@andygrunwald @ghostsquad Perhaps we should set up dependabot so that it can auto-pr package upgrades and merge if the CI passes?

andygrunwald:
<img width="965" alt="Screen Shot 2021-02-08 at 16 24 16" src="https://user-images.githubusercontent.com/320064/107240380-219c6380-6a2a-11eb-82a8-775ef1d48375.png">

Everything is activated already. Is there anything else to do from our side?

Is there a way to check if this vuln is already caught up by the bot?

benjivesterby:
Hmm I'm not sure. Do you have it setup here? https://app.dependabot.com/accounts/andygrunwald

andygrunwald:
Thanks @benjivesterby 
Dependabot is now working. PRs are incoming. Until now, no one for go-jwt.
Maybe we wait a bit and/or do it by hand :)

benjivesterby:
@andygrunwald no problem. I would suspect that it will auto-PR it. When I've had a security vuln pop on my libs before it took a couple days before dependabot updated. If not that we can keep an eye and do it manually.

lodi-g:
Thank you for your reactivity!
Please keep in mind that the newest version of jwt-go is a "preview" version so it might not be publised at the "latest" but more in a beta form, we should make sure it does not break anything before upgrading.

yarlson:
@andygrunwald https://github.com/dgrijalva/jwt-go has been abondoned (see https://github.com/dgrijalva/jwt-go/issues/462). In order to fix the vulnarability we have to switch to a community driven fork https://github.com/golang-jwt/jwt . The issue has been fixed in https://github.com/golang-jwt/jwt/pull/12

PR #403, opened 2021-09-29
GH-400, GH-401 Upgrading JWT Library to V4

benjivesterby:
# Description

Upgraded JWT library to v4 which now has module support as well as resolves security issues listed in GH-400, and GH-401

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -11,7 +11,7 @@ jobs:
     name: Test and lint
     strategy:
       matrix:
-        go-version: [1.x, 1.15.x, 1.14.x]
+        go-version: [1.15.x, 1.16.x, 1.17.x]
         platform: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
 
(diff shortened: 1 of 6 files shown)


benjivesterby:
@andygrunwald @ghostsquad Looks like this upgrade may break pre-1.16 support. Are we ok with that?

andygrunwald:
Also related:
* https://github.com/andygrunwald/go-jira/pull/382
* https://github.com/andygrunwald/go-jira/issues/343

andygrunwald (approved):
For me, breaking compatibility before go 1.15 is ok. Mainly because the current supported go versions are 1.15, 1.16, and 1.17.
We should follow the release cycle for the language.

The PR looks good. Thanks for taking care. I will move on and merge it.

About the `request_legacy.go` and `request_context.go`, we should think about dropping it at some point.

sitikanthasezzle:
can anybody update new tags or release version to v1.15.0

sitikanthasezzle:
@benjivesterby @andygrunwald

PR #406, opened 2021-10-06
Add request body logger

alexsuslov:
Hi,

Jira return additional information in body response. Let's add logger for error response body. 
To activate, you need to add the environment variable 
```
export DEBUG_REQUEST = YES
```
BRGD
Alex

### DIFF
--- a/error_logger.go
+++ b/error_logger.go
@@ -1,0 +1,16 @@
+package jira
+
+import (
+	"io"
+	"log"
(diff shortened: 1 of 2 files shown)


benjivesterby (changes requested):
I think this needs to be done differently. As a library accessing environment variables doesn't make sense. Can you please refactor and provide a different solution?

PR #428, opened 2021-12-22
Missing alias for import in example in README

tstrijdhorst:
Tried to run this example but it wouldn't run without specifying the alias.


benjivesterby:
Good catch. I'll merge after the tests pass.

benjivesterby (approved):

PR #430, opened 2021-12-29
add missing min/max to BoardConfigurationColumn

brenol:
# Description

Fixing missing fields in BoardConfigurationColumn struct. 

So we can identify what is the min/max of a column and to complete the API implementation 👍 
Doc URL: https://developer.atlassian.com/cloud/jira/software/rest/api-group-board/#api-agile-1-0-board-boardid-configuration-get

Doc pic:
image

I think it's a new feature, but its also a bug fix, so I labeled the commit as a feature.

Backwards compatible

Did not find any issue

Unfortunately I'm not sure.

## Example:

```go
// Example code
id := 26
boardConf, _, err := jira.Board.GetBoardConfiguration(id)
if err != nil { 
    panic(err)
}
// assuming your board has columns...
log.Println(boardConf.ColumnConfig.Columns[0].Min)
log.Println(boardConf.ColumnConfig.Columns[0].Max)
```

# Checklist

* [X] Unit or Integration tests added
  * [X] Good Path
  * [ ] Error Path
* [X] Commits follow conventions described here:
  * [X] [Conventional Commits 1.0.0](https://conventionalcommits.org/en/v1.0.0-beta.4/#summary)
  * [X] [The seven rules of a great Git commit message](https://chris.beams.io/posts/git-commit/#seven-rules)
* [X] Commits are squashed such that
  * [X] There is 1 commit per isolated change
* [X] I've not made extraneous commits/changes that are unrelated to my change.

### DIFF
--- a/board.go
+++ b/board.go
@@ -117,6 +117,8 @@ type BoardConfigurationColumnConfig struct {
 type BoardConfigurationColumn struct {
 	Name   string                           `json:"name"`
 	Status []BoardConfigurationColumnStatus `json:"statuses"`
+	Min    int                              `json:"min"`
+	Max    int                              `json:"max"`
 }
 
 // BoardConfigurationColumnStatus represents a status in the column configuration
--- a/board.go
+++ b/board.go
@@ -117,8 +117,8 @@ type BoardConfigurationColumnConfig struct {
 type BoardConfigurationColumn struct {
 	Name   string                           `json:"name"`
 	Status []BoardConfigurationColumnStatus `json:"statuses"`
-	Min    int                              `json:"min"`
-	Max    int                              `json:"max"`
+	Min    int                              `json:"min,omitempty"`
+	Max    int                              `json:"max,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


board.go
```diff
@@ -117,6 +117,8 @@ type BoardConfigurationColumnConfig struct {
 type BoardConfigurationColumn struct {
 	Name   string                           `json:"name"`
 	Status []BoardConfigurationColumnStatus `json:"statuses"`
+	Min    int                              `json:"min"`
```
benjivesterby, line 120: Please add the `omitempty` tag to the json config for these new fields as well as tests with omitted min/max

benjivesterby (changes requested):

board.go
brenol: Thank you for the quick reply! Made the changes. 😀

benjivesterby (approved):

andygrunwald (approved):

andygrunwald:
Thanks @brenol (and @benjivesterby for the review!)

Issue #417, opened 2021-11-13
How whe can get custom fields values?


erdvillegasxpt:
Hi team,

Recently we are creating a custom app that consume JIRA data, whe need to get values trhought custom fields, can you help me with an example of how can I read the values from that fields.

I using thise example:

```
func crearCSV(issues []jira.Issue) {

	file, err := os.Create("result.csv")
	if err != nil {
		panic(err)
	}
	defer file.Close()

	w := csv.NewWriter(file)
	w.Comma = ','
	defer w.Flush()

	w.Write([]string{"IssueKey,IssueType,ProjectKey,Summary"})
	checkError(err)

	for _, i := range issues {

		a, _ := i.Fields.Unknowns.Value("customfield_10061")
		fmt.Printf("%#v", a)

	}

}
```

evilnerd:
The thing you're looking for is `jira.Issue.GetCustomFields(issueKey)` - this returns a map with custom field values where the key is the field's internal name. 

If you want to map the 'user friendly' name to the internal name, I recommend you use `jira.Client.Field.GetList()`, to build a 'translation' map (which is also safer, since the internal name is not guaranteed to be the same if you have e.g. a test and prod version of Jira)

```go
	fields, _, err := jira.Client.Field.GetList()
	if err != nil {
		return fmt.Errorf("unable to get the list of fields: %w", err)
	}

	jira.fieldMapping = make(map[string]string)

	for _, f := range fields {
		if f.Custom {
			jira.fieldMapping[f.Name] = f.ID
		}
	}

```

erdvillegasxpt:
Thanks for the example, as I try it, I let you know the results. Thanks a lot!

erdvillegasxpt:
Thanks for your help

andygrunwald:
@erdvillegasxpt Would you be able to prepare a pull request to extend the docs?

erdvillegasxpt:
Sure, let me write the note and I'll send a PR

andygrunwald:
Hey,

I am very sorry that this issue has been open for a long time with no final solution. We work on this project in our spare time, and sometimes, other priorities take over. This is the typical open source dilemma.

#### However, there is news: **We are kicking off v2 of this library 🚀** 

To provide visibility, we created the [Road to v2](https://github.com/andygrunwald/go-jira/milestone/1) Milestone and calling for your feedback in https://github.com/andygrunwald/go-jira/issues/489

[...]

Issue #380, opened 2021-06-10
Custom fields not being written, and creating issues returns an incorrect *jira.Issue with nil fields


modernplumbing:
## Describe the bug
I'm on Jira Cloud trying to create new issues with custom fields. I have the following code, and the custom field is not getting written. Additionally, the createdIssue has a nil value for Fields, even though in the UI, the other specified fields minus the custom fields are getting written.
```
		newIssue := jira.Issue{
			Fields: &jira.IssueFields{
				Summary:     oldIssue.Fields.Summary,
				Description: oldIssue.Fields.Description,
				Project: jira.Project{
					Key: "NEWBOARD",
				},
				Type: jira.IssueType{
					Name: "NEWTYPE",
				},
				Assignee: oldIssue.Fields.Assignee,
				Duedate:  oldIssue.Fields.Duedate,
				Unknowns: tcontainer.MarshalMap{
					"customfield_11228": 1386,
				},
			},
		}

createdIssue, resp, err := c.JClient.Issue.Create(&newIssue)
```

## To Reproduce
Adding 
```Unknowns: tcontainer.MarshalMap{
	"customfield_11228": 1386,
},
``` 
to https://sourcegraph.com/github.com/andygrunwald/go-jira/-/blob/examples/createwithcustomfields/main.go should be reproduceable. 

## Expected behavior

I expect this custom field to be written and visible in the UI. I also expect the returned *jira.Issue (createdIssue in the above example) to reflect the issue that was created, i.e. not having a nil value for the Fields.

## Your Environment

Jira Cloud 
github.com/andygrunwald/go-jira v1.13.0
github.com/trivago/tgo v1.0.7

## Additional context

I am migrating tickets from one board to another. Because of the custom fields, I need to create new issues in the new project, but I am blocked on being unable to write the custom fields.

andygrunwald:
Hey @modernplumbing,
thanks for the issue. I am curious, why do you close it already?
Were you able to find a solution? If yes, would you mind sharing it with the community?

danekantner:
To add to the above, to set a labels field instead it's the same syntax as if it were just a string except you provide an array:

`	unknown.Set("customfield_11503", []string{"label1", "label2", "label3"})`

andygrunwald:
Hey,

I am very sorry that this issue has been open for a long time with no final solution. We work on this project in our spare time, and sometimes, other priorities take over. This is the typical open source dilemma.

[...]

PR #459, opened 2022-05-30
BREAKING CHANGE: Use create request body instead of custom parameters

Fank:
# Description

atm the use of custom fields is impossible, because the value could only be a string, so a field with type `option-with-child` which requires a `{ "value": "foo" }` as value wont work.
By allowing such change, the struct needed to changed so it would be a breaking change anyway.
While doing that i also removed all parameters which are a "part" of the struct and put them together.

Allowing to transfer more complex custom field values

Currently only the value string can be used, but more complex structure is needed to for custom fields.
And the request struct type which was given is not the same which is mentioned in the jira docs, so i changed the field structure to match it again.

New feature
Code quality improvements

Yes, no backward compatibility

Not directly, maybe:
- #417
- #380
   
Cloud

## Example:

```go
// Example code

```

# Checklist

  * [ ] Good Path
  * [ ] Error Path
  * [x] [Conventional Commits 1.0.0](https://conventionalcommits.org/en/v1.0.0-beta.4/#summary)
  * [x] [The seven rules of a great Git commit message](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/request.go
+++ b/request.go
@@ -3,6 +3,7 @@ package jira
 import (
 	"context"
 	"fmt"
+	"net/http"
 )
 
 // RequestService handles ServiceDesk customer requests for the Jira instance / API.
@@ -27,7 +28,7 @@ type Request struct {
 type RequestFieldValue struct {
 	FieldID string `json:"fieldId,omitempty" structs:"fieldId,omitempty"`
 	Label   string `json:"label,omitempty" structs:"label,omitempty"`
-	Value   string `json:"value,omitempty" structs:"value,omitempty"`
+	Value   any    `json:"value,omitempty" structs:"value,omitempty"`
 }
 
 // RequestDate is the date format used in requests.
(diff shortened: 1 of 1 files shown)


andygrunwald:
Hey,

I am very sorry that this pull request has been open for a long time with no final feedback or merge/decline. We work on this project in our spare time, and sometimes, other priorities take over. This is the typical open source dilemma.

#### However, there is news: **We are kicking off v2 of this library 🚀** 

To provide visibility, we created the [Road to v2](https://github.com/andygrunwald/go-jira/milestone/1) Milestone and calling for your feedback in https://github.com/andygrunwald/go-jira/issues/489

The development will take some time; however, I hope you can benefit from the changes.
If you seek priority development for your pull request + you like to sponsor it, please contact me.

#### What does this mean for my pull request?

We will work on this pull request indirectly.
We might merge it during the development or pull parts of it into the new version.
This means that during the development phase, we aim to tackle it.
Maybe in a different way like it is currently handled.
Please understand that this will take a while because we are running this in our spare time.

#### Final words

Thanks for using and contributing to this library.
If there is anything else you would like to tell us, let us know!

PR #500, opened 2022-08-30
ci: add Reviewpad action

marcelosousa:
# Description

This pull request adds the [Reviewpad action](https://github.com/reviewpad/action) to the repository to start automating workflows around pull requests without adding a new GitHub action to each automation.

The automation added are specified in the `.github/reviewpad.yml` file:

- Add a label depending on the size of the pull request;
- Add a label depending on the modifications (replaces the existing labeler action - FYI: that action is only active in pull requests which means that it wouldn't run on contributions from forks and the extra permissions are not necessary);
- Automatically writes a welcome message to new contributors to the project
- Checks if there the history of a pull request is linear; if it is not, then adds a warning in the pull request as a comment.

Would be great to get some feedback on this!


andygrunwald:
We won't add Reviewpad right now to this repository.
Thanks for the suggestion.

Issue #473, opened 2022-08-20
Split Jira OnPremise and Jira Cloud clients


andygrunwald:
## Context

The APIs between the two JIRA instances, On-premise and Cloud, differ from each other.

## Current behavior

The current implementation covers one client that implements a wild mix between OnPremise API endpoints and Cloud API endpoints.
This leads sometimes to bug reports and unexpected behavior because the API endpoints (URI) are equal, but the parameters differ.

## Breaking change

This is a breaking change, is planned for v2 and need to be documented.

andygrunwald:
* Jira Server REST API: https://developer.atlassian.com/server/jira/platform/rest-apis/
* Jira Cloud REST API v3: https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/

PR #503, opened 2022-09-07
Split On-Premise and Cloud Clients #473

andygrunwald:
The APIs between the two JIRA instances, On-premise and Cloud, differ from each other.
This is the first Pull Request that splits the clients between On-Premise and Cloud.

I decided to copy the clients, including the utility functions.
The main reasoning is:
- Changes are easier to test (only the affected client needs to be tested)
- Code can diverge without any issues
- The utility functions are not that much / This model is easier to keep inside the head

### TODO

- [x] Adjust changelog / migration guide

Related #473


Fank (approved):

Fank:
An explanation about how to handle cloud and onpremise should be added README.md, but can be done later because i think there will be more changes in the future.

andygrunwald:
@Fank Yeah, docs (mainly README) are a whole different story and will be done when we progress further.

PR #517, opened 2022-09-11
Get v2 working with go modules

### LINKED ISSUE #482: Get v2 working with go modules
## Context

See https://blog.golang.org/v2-go-modules

andygrunwald:
Fix #482


andygrunwald:
As far as I can see, only the examples are failing.
The examples itself are not able to download v2, because this is the part of this PR.

I will merge it through. Then, the examples should work again.

Issue #495, opened 2022-08-23
Support Jira Insight


Fank:
Jira bought "Insight" a longer time ago and added it to there Cloud APIs using the same authentication methods:

> Authentication: For authenticating directly against the REST API, the Insight in Jira Service Management REST API supports basic auth (see [Basic auth for REST APIs](https://developer.atlassian.com/cloud/confluence/basic-auth-for-rest-apis/)).
Reference https://developer.atlassian.com/cloud/insight/intro/introduction-and-basics/

## Additional context

REST Documentation: https://developer.atlassian.com/cloud/insight/rest/

andygrunwald:
See https://gitlab.com/mclgmbh/golang-pkg/jira-insight/-/blob/main/insight/api.go

Fank:
@andygrunwald While implementing #565 i noticed that all endpoints require a workspace id.
So I came up with some ideas right now, should we go for:

1.
```go
client := NewClient(...)
client.Insight.ObjectType.GetAttributes("MyWorkspaceID", "ID")
```

2.
```go
client := NewClient(...)
client.Insight("MyWorkspaceID").ObjectType.GetAttributes("ID")
```

3.
```go
client := NewClient(...)
client.Insight.ObjectType.GetAttributes("MyWorkspaceID", "ID")

client2 := NewClient(...)
client2.InsightWithWorkspace("MyWorkspaceID").ObjectType.GetAttributes("ID")
// which is a wrapper for
// client2.Insight.ObjectType.GetAttributes("MyWorkspaceID", "ID")
// so you can do something like:

insight := client2.InsightWithWorkspace("MyWorkspaceID")
insight.ObjectType.Get("ID")
insight.ObjectType.GetAttributes("ID")
...
```

andygrunwald:
Personally, I am in favor of option 1, because

* it is pretty easy to understand from a caller perspective
* the library might end up being more simple due to no logic to "hold" the workspace id
* it fits other areas of the library (e.g., where kanban or sprint id needs to be added)

PR #565, opened 2022-09-12
Added initial insights

Fank:
#### What type of PR is this?

feature

Adds cloud app insight: https://developer.atlassian.com/cloud/insight/rest/

Relates to #495

Missing ToDos:
- [ ] ~~Added documentation~~ (will be added later)
- [x] Added tests


cloud/insight_object.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	req.Header.Set("Accept", "application/json")
+
+	res, err := i.client.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	switch res.StatusCode {
```
Fank, line 62: The insight api uses the same response handling as jira api, so the idea was to make it more lightweight and easier to handle

cloud/models/apps/insights/basic.go
```diff
@@ -0,0 +1,54 @@
+package insights
```
Fank, line 1: The idea behind this is to separate the models from the request module to track versions, in this case its a little bit harder because the versions are 6-8 atm.
But for the other cases it would be:
- cloud/models/v2
- cloud/models/v3
- cloud/models/v4
- onpremise/models/v2
- .....

andygrunwald:
Thanks for this Pull Request. Overall it looks quite good.
I would like to run through a few points with you

## `model` package

What is the thought behind the own model package?
The package contains all structs in the top-level package right now.

Are there name conflicts?

## Insight API vs. single services

In the current implementation (for [object Get objecttype {id} attributes](https://developer.atlassian.com/cloud/insight/rest/api-group-objecttype/#api-objecttype-id-attributes-get)), you call it like

```go
client.Insight.GetObjectTypeAttributes(...)
```

What do you think about an API like

```go
client.Insight.Objecttype.GetAttributes(...)
```

The sublevel (here Objecttype) would be in line with the documentation navigation structure at https://developer.atlassian.com/cloud/insight/rest/api-group-objecttype/#api-objecttype-id-attributes-get

## Unit tests

It would be nice to have a few (you listed those as open TODOs already)

## Checking of status codes

The code

```go
switch res.StatusCode {
case http.StatusUnauthorized:
	return nil, fmt.Errorf("%s: %w", req.URL.String(), ErrUnauthorized)
case http.StatusNotFound:
	return nil, fmt.Errorf("%s: %w", req.URL.String(), ErrNotFound)
case http.StatusInternalServerError:
	return nil, fmt.Errorf("%s: %w", req.URL.String(), ErrUnknown)
}
```

It looks like it is nearly all the same.
Would we be able to make this more generic and place this in the Do method (where the request is executed)?

Fank:
> ## `model` package
> What is the thought behind the own model package? The package contains all structs in the top-level package right now.
> 
> Are there name conflicts?

No there are not (yet), buts its an app from the [marketplace for the cloud service](https://marketplace.atlassian.com/apps/1212137/insight-asset-management?hosting=datacenter&tab=overview) and because of that i would like to separate it because its tracks a completely different version and naming.
So it easier to find out where are those structs are coming from and why are they there.


> ## Insight API vs. single services
> In the current implementation (for [object Get objecttype {id} attributes](https://developer.atlassian.com/cloud/insight/rest/api-group-objecttype/#api-objecttype-id-attributes-get)), you call it like
> 
> ```go
> client.Insight.GetObjectTypeAttributes(...)
> ```
> 
> What do you think about an API like
> 
> ```go
> client.Insight.Objecttype.GetAttributes(...)
> ```
> 
> The sublevel (here Objecttype) would be in line with the documentation navigation structure at https://developer.atlassian.com/cloud/insight/rest/api-group-objecttype/#api-objecttype-id-attributes-get

Yes i totally missed that, its changed as written above.

> ## Checking of status codes
> The code
> 
> ```go
> switch res.StatusCode {
> case http.StatusUnauthorized:
> 	return nil, fmt.Errorf("%s: %w", req.URL.String(), ErrUnauthorized)
> case http.StatusNotFound:
> 	return nil, fmt.Errorf("%s: %w", req.URL.String(), ErrNotFound)
> case http.StatusInternalServerError:
> 	return nil, fmt.Errorf("%s: %w", req.URL.String(), ErrUnknown)
> }
> ```
> 
> It looks like it is nearly all the same. Would we be able to make this more generic and place this in the Do method (where the request is executed)?

This is a proof of concept to check if this is a solid way to handle errors, so far it looks good.
I would like to replace the existing checks for bad status code with this behavior in general, because i see the same status codes in Jira REST and Service Desk too.
Is this fine to replace the current status code handling with this attempt?
If yes i would like to separate it into a new PR

Fank:
> ## Unit tests
> It would be nice to have a few (you listed those as open TODOs already)
> 

Added

Fank:
This is the first batch and contains all functions i need at the moment.

Fank:
> ## Checking of status codes
> The code
> 
> ```go
> switch res.StatusCode {
> case http.StatusUnauthorized:
> 	return nil, fmt.Errorf("%s: %w", req.URL.String(), ErrUnauthorized)
> case http.StatusNotFound:
> 	return nil, fmt.Errorf("%s: %w", req.URL.String(), ErrNotFound)
> case http.StatusInternalServerError:
> 	return nil, fmt.Errorf("%s: %w", req.URL.String(), ErrUnknown)
> }
> ```
> 
> It looks like it is nearly all the same. Would we be able to make this more generic and place this in the Do method (where the request is executed)?

Will be changed with #570 , i would keep the current handling and replace it in the linked PR.

andygrunwald:
Thanks. This PR is still on my list to review and move forward. It might take a few days.

ChristianKniep:
Hey there, what is the status of the asset integration. I am going to need that soon(-ish), so happy to help.

Fank:
@ChristianKniep this PR is abandoned
Currently working on that here https://github.com/ctreminiom/go-atlassian/pull/204

andygrunwald:
Just Reopening this.
I think there are a few good ideas in this PR.
Right now, my time is pretty limited, still, I plan to come back to this.

Issue #559, opened 2022-09-11
Review service "ServiceDesk" in the Cloud Jira client


andygrunwald:
In #473 we split the single Jira client into two: OnPremise and Cloud.
We did this by copying the same code into two folders. The reason: The two software products have different APIs.
In v1 of this client, we had a bad/dirty mix of OnPremise and Cloud API endpoints.

## Goal of this task

Review service "ServiceDesk" in the Cloud Jira client and

1. delete all API endpoints related to the OnPremise Jira version
2. update the documentation to match the latest API links

## Notes

This might reduce the functionality of the client, but this lead to a more stable client.

Missing API endpoints will be documented that we have a clear overview.

Fank:
I would recommend based on your [idea](https://github.com/andygrunwald/go-jira/pull/565#issuecomment-1244244545) to move all service desk related APIs into service desk e.g.
```go
client.ServiceDesk.Request.Create(...)
client.ServiceDesk.RequestType.Get(...)
client.ServiceDesk.Customer.Create(...)
```

andygrunwald:
From the first quick check: Sounds like a good idea.
Depending on the amount of items under each sub category, it might also be a idea to consider to run it like

```go
client.ServiceDesk.CreateRequest(...)
client.ServiceDesk.GetRequestType(...)
client.ServiceDesk.CreateCustomer(...)
```

Just a quick thought.
In the end, this should be consistent across the library.

PR #458, opened 2022-05-25
Added requesttype

Fank:
# Description

For our current implementation against jira we decided to use this module, but it lacks some requests we need which we would like to contribute in this and upcoming pull requests.
This PR will only focus on the [Requesttype](https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-requesttype/).
I noticed that there is currently a `PagedDTO` struct which contains `[]ìnterface{}` for `values`.
To improve the handing i decided to support this new Requesttype for only golang 1.18+ and use a new struct `PagedDTOT` which is a generic version of the normal one `PagedDTO`.

Information that is useful here:
* **The What**: What is your change doing?

Adds support for non experimental `requesttype` using `serviceDeskID`
https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-requesttype/

* **The Why**: Why is your change useful/needed? What is your use case? (this helps us to understand the real world better)

Currently there is no support for that in this module

* **Type of change**: Things like Bugfix, New feature, Code quality improvements, Dependency upgrade, Documentation, ...

Bugfix
New feature
Code quality improvements

* **Breaking change**: Yes or no? Backward compatible?

No breaking change

* **Related to an issue**: Does this fix or close an issue? Or is related in any kind?

Nope

* **Jira Version + Type**: Which Jira version and type (on-premise / cloud) you have used?

## Example:

Let us know how users can use or test this functionality.

```go
// Example code

jiraClient, err := jira.NewClient(nil, "jiraurl")
if err != nil {
  return err
}

pager, _, err := jiraClient.RequestType.GetWithContext(ctx, 1)
if err != nil {
  return err
}

fmt.Println(pager.Size)

for _, v := range pager.Values {
  fmt.Println(v.Name)

  fmt.Println(v.Icon.Links.IconURLs)
}
```

# Checklist

* [ ] Unit or Integration tests added
  * [ ] Good Path
  * [ ] Error Path
* [x] Commits follow conventions described here:
  * [x] [Conventional Commits 1.0.0](https://conventionalcommits.org/en/v1.0.0-beta.4/#summary)
  * [x] [The seven rules of a great Git commit message](https://chris.beams.io/posts/git-commit/#seven-rules)
* [x] Commits are squashed such that
  * [x] There is 1 commit per isolated change
* [x] I've not made extraneous commits/changes that are unrelated to my change.

PR #566, opened 2022-09-12
Backport of all changes related to servicedesk

Fank:
#### What type of PR is this?

* feature
* api-change

Its a backport from all related changes in servicedesk from https://github.com/mcl-de/go-jira/tree/fk_internal

Related to #459, #458, #559, #560 & #561

Servicedesk API is a different API module compared to the Jira REST API, so i changed the naming of some of the files with a "servicedesk_" prefix and moved the models into a subfolder called servicedesk.
Not sure about the models subfolder yet, if the names are always different and if similar in name and properties this subfolder can be merged into the models folder.
But otherwise its way harder to split them again if we notice a difference.


Fank:
I would check tests and other stuff which is missing when doing https://github.com/andygrunwald/go-jira/issues/559#issuecomment-1245266085

andygrunwald:
Thanks for this Pull Request. It is still on my list to review. It might take a few days, thanks for understanding.

PR #570, opened 2022-09-13
Generic response error handling

Fank:
#### What type of PR is this?

* cleanup
* feature
* api-change

This PR replaces the existing request error handling with a more generic variation which allows it to extract the response body depending on the use case and handle the status code which isn't possible at the moment.

Relates to https://github.com/andygrunwald/go-jira/pull/565#issuecomment-1244244545

I am not familiar with any kind of error messages, so far i only checked the insight API and there errors messages are not static or being able to parse to static structs.

The idea was to have use cases like:
```go
issue, _, err := client.Issue.Get(context.Background(), "123", nil)
if err != nil && errors.Is(err, ErrNotFound) {
	// create new issue
} else err != nil {
	return err
}

issue, _, err := client.Issue.Create(context.Background(), &Issue{})
if err != nil && errors.Is(err, ErrValidation) {
	// check error content and proceed as preferred
} else err != nil {
	return err
}
```

### DIFF
--- a/cloud/board.go
+++ b/cloud/board.go
@@ -142,8 +142,7 @@ func (s *BoardService) GetAllBoards(ctx context.Context, opt *BoardListOptions) 
 	boards := new(BoardsList)
 	resp, err := s.client.Do(req, boards)
 	if err != nil {
-		jerr := NewJiraError(resp, err)
-		return nil, resp, jerr
+		return nil, resp, err
 	}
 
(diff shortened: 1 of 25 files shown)


andygrunwald:
Thanks for this PR. I might take some time to review this, but it is still on my list.

One quick question:
Do you have any particular things in mind that need to be done before this PR can move out of the Draft state?

PR #573, opened 2022-09-13
Replaced ReadAll and Unmarshal with NewDecoder

Fank:
#### What type of PR is this?

* cleanup

Replaced all `io.ReadAll` & `json.Unmarshal` with `json.NewDecoder(...).Decode(...)`.
In some cases the original error was not returned, which is now the case.

Was found while changing code in #570

### DIFF
--- a/cloud/authentication.go
+++ b/cloud/authentication.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 )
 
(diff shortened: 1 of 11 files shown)


andygrunwald:
Thanks a lot @Fank!
I did a few (minor) follow up changes:
* https://github.com/andygrunwald/go-jira/commit/469abf12a507317f276fd0c17b65b62550c0c76f
* https://github.com/andygrunwald/go-jira/commit/75eddf5684f34e4d6dfaf6a5e5931d018b7f10cb
* https://github.com/andygrunwald/go-jira/commit/221b42e8ceb2df3fb94c2f0b72f88a78fe2fbbf3

PR #591, opened 2022-10-27
Added issue link

Fank:
#### What type of PR is this?

* feature

It adds the missing [issue link](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-links/#api-group-issue-links)

### DIFF
--- a/cloud/issuelink.go
+++ b/cloud/issuelink.go
@@ -1,0 +1,105 @@
+package cloud
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+// IssueLinkService handles issue links for the Jira instance / API.
+//
+// Jira API docs: https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-links/#api-group-issue-links
+type IssueLinkService service
+
+type CreateLink struct {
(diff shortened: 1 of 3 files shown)


andygrunwald:
Thanks for this Pull Request @Fank.
This is looking pretty good already!

A few things, that came into my mind:

### Own Service
I am not 100% sure if we should run our own service here or consider making this part of the IssueService.
IssueLinks are pretty coupled to Issues.
And the IssueLinkService would be pretty small (only 3 methods).
I don't have a correct answer for me right now, just a thought.

### Existing methods
In `issue.go`, we have already a few IssueLink API methods:

* https://github.com/andygrunwald/go-jira/blob/aea7aec6b799e6f614153a8e89939726648a4098/cloud/issue.go#L735
* https://github.com/andygrunwald/go-jira/blob/aea7aec6b799e6f614153a8e89939726648a4098/cloud/issue.go#L1022

They are (still) a wild mix between On-Premise and Cloud, but I would consider removing those already.

### Moving forward
Let me know your thoughts.
I can make the changes myself, while merging this into in the next couple of time/days.

PR #593, opened 2022-11-23
fix(code-snippet): example jira ticket status update

roshanraj:
var testIssueID was not declared but was being used across the example.

#### What type of PR is this?
documentation, code snippet fix

Use of undeclared variable in example, may cause confusion among potential developers

#594 
Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,7 @@ import (
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


andygrunwald (approved):
Thank you

Issue #390, opened 2021-07-26
Expand field from search isn't parsed


leviska:
## Is your feature request related to a problem? Please describe.

You can add "expand" field to the search request  
https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-get  
https://pkg.go.dev/github.com/andygrunwald/go-jira?utm_source=godoc#SearchOptions  
So if you add "names" or "schema" for example, you can't get them from the result, because library will read the body, and will not parse this fields

## Describe the solution you'd like

Add this fields into [searchResult](https://github.com/andygrunwald/go-jira/blob/master/issue.go#L515) and somehow return them from Search function

## Describe alternatives you've considered

I can make the request myself, but it's strange, that you can add "expand" in request, but can't read the result

leviska:
If you add "expand: names", you'll have something like this in response:
```json
{
    "expand": "schema,names",
    "startAt": 0,
    "maxResults": 50,
    "total": 7,
    "issues": [],
    "names": {}
}
```

And as you can see from `searchResult`, the "names" field isn't parsed

andygrunwald:
Hey,

I am very sorry that this issue has been open for a long time with no final solution. We work on this project in our spare time, and sometimes, other priorities take over. This is the typical open source dilemma.

#### However, there is news: **We are kicking off v2 of this library 🚀** 

To provide visibility, we created the [Road to v2](https://github.com/andygrunwald/go-jira/milestone/1) Milestone and calling for your feedback in https://github.com/andygrunwald/go-jira/issues/489

The development will take some time; however, I hope you can benefit from the changes.
If you seek priority development for your issue + you like to sponsor it, please contact me.

#### What does this mean for my issue?

We will work on this issue indirectly.
This means that during the development phase, we aim to tackle it.
Maybe in a different way like it is currently handled.
Please understand that this will take a while because we are running this in our spare time.

#### Final words

Thanks for using this library.
If there is anything else you would like to tell us, let us know!

PR #601, opened 2022-12-16
Added POST search & updated GET search for latest api version

### LINKED ISSUE #395: Support for Jira API v3
## Is your feature request related to a problem? Please describe.

I am developing a project dependent on the go-jira and I would like to start by using v3 Jira APIs (https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/)

Currently, (almost) all API calls are based on the v2 Jira API version. I can (and have) use the NewRequest/Do pattern described in the examples, however this is kludgy. I would prefer to use direct methods to obtain the results in the same manner that v2 API is currently supported. Also, since the v3 structures are not supported then I would be required to re-implement the structures in my project.

## Describe the solution you'd like

I would like to have built-in support for v3 APIs. Preferably so that both v2 and v3 are supported by the same version of the library.

## Describe alternatives you've considered

As mentioned in the description I can request the v3 APIs directly through NewRequest/Do and write my own deserializer for the changed structures but this is suboptimal as I am probably duplicating work that others require as well :)

## Additional context

Looking at the OpenAPI specification of the v3 APIs there are not that many changes, although that may depend on your point of view. For the most part, the v2 and v3 just have different endpoints (replacing 2 with 3, surprisingly :) ), but otherwise they return almost the same structures.

The major difference is the ADF (Atlassian Document Format) which impacts how some fields are constructed. Instead of strings some APIs return values describing documents in specific JSON format. Please see the above link to the v3 API for the details. Additionally, note that while testing I have also seen some differences with parameters' support between the v2 and v3 APIs (specifically, the difference between `project` and `project/search`) but, while sometimes confusing, this can be addressed by samples and documentation.

Fank:
#### What type of PR is this?

* feature

Added POST search function in cloud.
General cleanup of the issue search in cloud.

Fixes #395
Related #390

The search returns now the original json as struct instead of issue array and the issue array is untouched, marked as TODO for later fixes because it uses the wrong type as defined in documentation.
JQL is now inside the options block because both GET & POST are using this for building the URL as for JSON.

### DIFF
--- a/cloud/issue.go
+++ b/cloud/issue.go
@@ -1123,29 +1123,22 @@ func (s *IssueService) SearchPOST(ctx context.Context, options *SearchOptions) (
 //
 // TODO Double check this method if this works as expected, is using the latest API and the response is complete
 // This double check effort is done for v2 - Remove this two lines if this is completed.
-func (s *IssueService) SearchPages(ctx context.Context, jql string, options *SearchOptions, f func(Issue) error) error {
-	if options == nil {
-		options = &SearchOptions{
-			StartAt:    0,
-			MaxResults: 50,
-		}
-	}
-
+func (s *IssueService) SearchPages(ctx context.Context, options *SearchOptions, f func(Issue) error) error {
 	if options.MaxResults == 0 {
 		options.MaxResults = 50
 	}
 
-	issues, resp, err := s.Search(ctx, jql, options)
+	search, resp, err := s.SearchPOST(ctx, options)
 	if err != nil {
 		return err
 	}
 
-	if len(issues) == 0 {
+	if len(search.Issues) == 0 {
 		return nil
 	}
 
 	for {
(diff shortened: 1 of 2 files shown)


Fank:
One of the main issues i was running into is that a major change between current v2 and new v3 in the API version is the issue field `description` which changed from a `string` into an unknown type.

`v2`:
```
"description": "dsffdsfd",
```

`v3`:
```
"description": {
  "version": 1,
  "type": "doc",
  "content": [
    {
      "type": "paragraph",
      "content": [
        {
          "type": "text",
          "text": "dsffdsfd"
        }
      ]
    }
  ]
},
```

Which causes a lot of issues with the struct `IssueFields` because it needs to be changed in all occurrences.

andygrunwald:
Thanks @Fank!
I will check it out in the next days. Happy Xmas

PR #627, opened 2023-05-30
GetList Unmarshalling Fix

anwalsh:
#### What type of PR is this?

* bug

Resolves an issue with `GetList` behavior in which the JSON payload could not be properly unmarshalled into the desired data object. Before it would fail for users of the API as the file had closed while unmarshalling. This presents as being due to the returned JSON payload being a document containing a JSON property with a array value containing the subdocuments we care about and not a top level array of the documents we care about.

I've made the necessary adjustments, all tests pass. 

Let me know if anything else should change.

Note: I removed the two TODO comments as they both function as expected _now_. Not sure if that is correct? I'd also like to assist with the march to V2 more directly if possible, not sure if there is a communication channel for coordination/guidance?

Thank you!


andygrunwald:
Thanks @anwalsh for this pr.

I wonder, why was this closed?

anwalsh:
@andygrunwald I ended up scrapping this PR in favor of a clean one - apologies there! I made a weird mistake with git by modifying the git history on this PR. In order to fix that I reset everything to draft the other PR for safety. 

Sorry for the confusion!

PR #630, opened 2023-05-31
Fix for GetList Unmarshalling Issue

anwalsh:
Currently the GetList() function for IssueLinkTypes in OnPremise and Cloud fails to return the JSON document from the API. This is because it expects a top level array of subdocuments and not a top level document containing an array of subdocuments. I have corrected this behavior and revised the tests to ensure the behavior is correct. I have also tested this behavior against my own Jira OnPremise and Cloud instances.

#### What type of PR is this?

* bug

As currently implemented GetList() does not return the expected list of `IssueLinkType`s. Instead, the function call returns an empty list and an error if probed deeply enough. All tests pass against the unit tests which further cause confusion. However, directly calling the endpoint and verifying the output we can discern that the function expects a different JSON document shape than that which is returned (see the test mock of the output). This PR addresses the above behavior for tests, onpremise, and Cloud. It also hopes to verify the v2 push for each of these functions - but I'd like feedback on that from a Maintainer before I call it "good".

NA - I didn't file an issue - apologies if that is incorrect, I'll be sure to redress that if we prefer that method.

Fixes #
OnPremise and Cloud unmarshalling for IssueLinkTypes array.

I'd like to be more involved with the project if I may. However, I'd likely need some guidance especially around the goals/nongoals specifically to do with the push for v2 and vision. I was wondering if there might be any mentorship ability or a channel/communication medium for one to leverage for ad hoc questions to assist in ramping/working on the project?

I had a previous PR up but I did something odd with git while trying to clean up my commits - apologies for that error.

Thank you!

### DIFF
--- a/cloud/issuelinktype.go
+++ b/cloud/issuelinktype.go
@@ -17,7 +17,7 @@ type IssueLinkTypeService service
 // Jira API docs: https://developer.atlassian.com/cloud/jira/platform/rest/v2/#api-rest-api-2-issueLinkType-get
 func (s *IssueLinkTypeService) GetList(ctx context.Context) ([]IssueLinkType, *Response, error) {
 	apiEndpoint := "rest/api/2/issueLinkType"
-	req, err := s.client.NewRequest(ctx, "GET", apiEndpoint, nil)
+	req, err := s.client.NewRequest(ctx, http.MethodGet, apiEndpoint, nil)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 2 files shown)


andygrunwald:
Thanks @prarit for this pr.
I am currently short on time for open-source contributions due to other/private obligations. This means the review of this PR can take a bit. Thanks for contributing to open source and thanks for understanding.

andygrunwald:
Thanks @anwalsh  for this pr.
I am currently short on time for open-source contributions due to other/private obligations. This means the review of this PR can take a bit. Thanks for contributing to open source and thanks for understanding.

anwalsh:
@andygrunwald no worries! I have something to fix it appears as the checks failed so I will take a look and see what I've messed up and get that sorted!

PR #649, opened 2023-12-04
feat(issue): add statuscategorychangedate

daveoy:
adding the StatusCategoryChangeDate field to the IssueFields struct which should marshall issue.fields.statuscategorychangedate from the jira cloud REST api response JSON to a time.Time representing the time an issue changed to its current status.


daveoy:
still testing

PR #650, opened 2023-12-06
feat(issue): add statuscategorychangedate

daveoy:
#### What type of PR is this?

* feature

Open for discussion, but this field is present on REST API responses from atlassian inside my organization's tenant. It represents the time a n issue has been present in its current status and I have based some metrics on calculations from the reported time.

Understand if you don't want to merge given this field appears undocumented in the REST API. Would love to know if its present in your organization as well (atlassian may need to update their docs in this case).

api docs here: https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issues/#api-rest-api-3-issue-issueidorkey-get

### DIFF
--- a/cloud/issue.go
+++ b/cloud/issue.go
@@ -121,6 +121,7 @@ type IssueFields struct {
 	Reporter                      *User             `json:"reporter,omitempty" structs:"reporter,omitempty"`
 	Components                    []*Component      `json:"components,omitempty" structs:"components,omitempty"`
 	Status                        *Status           `json:"status,omitempty" structs:"status,omitempty"`
+	StatusCategoryChangeDate      Time              `json:"statuscategorychangedate,omitempty" structs:"statuscategorychangedate,omitempty"`
 	Progress                      *Progress         `json:"progress,omitempty" structs:"progress,omitempty"`
 	AggregateProgress             *Progress         `json:"aggregateprogress,omitempty" structs:"aggregateprogress,omitempty"`
 	TimeTracking                  *TimeTracking     `json:"timetracking,omitempty" structs:"timetracking,omitempty"`


andygrunwald:
@aechweeks The [ProjectCategoryChangeEvent](https://docs.atlassian.com/software/jira/docs/api/7.2.2/com/atlassian/jira/event/project/ProjectCategoryChangeEvent.html) seems to be in the OnPremise version. The PR is changing the cloud version of the API here.

Thanks for the community post as well. Still no mention of cloud or on premise :( 

I will wait for someone from the community to know, if someone else sees this field.

aechweeks:
i can confirm it exists in the cloud version. is there a test i can provide you or a repo i can setup to prove it? this is def there and is the best way to check if something has or hasn't happened. 

the status show [here](https://docs.atlassian.com/software/jira/docs/api/7.7.2/com/atlassian/jira/event/project/class-use/ProjectCategoryChangeEvent.html), which indicates it can be used [here](https://docs.atlassian.com/software/jira/docs/api/7.7.2/com/atlassian/jira/event/project/package-summary.html) as an event. 

for more proof:

- [JQL using statusCategoryChangedDate](https://community.atlassian.com/t5/Jira-Service-Management/JIRA-JQL-Query-to-find-issues-statusCategoryChangedDate-after/qaq-p/2363293)
- [JQL Rec for using the key](https://confluence.atlassian.com/jirakb/how-to-search-using-statuscategory-statuscategorychangedate-function-with-jql-1115669437.html)

*statusCategoryChangeDate()*
 Similar to StatusCategory, you can issue based on the timeline using the statusCategoryChangedDate() function

andygrunwald:
Thanks all.

aechweeks:
thank you! i think your burden of proof is reasonable and leads to a higher quality package.

PR #655, opened 2024-01-27
Add FindAssignable method to User

zbarahal:
Feature: Adding a new method to the User interface

Adding support for the find assignable users route https://docs.atlassian.com/software/jira/docs/api/REST/1000.1568.0/#api/2/user-findAssignableUsers

No linked issue - can create if that's required.

I attempted to follow the pattern from User.Find by using the option funcs. The other option I considered was using the query library that's being used in Issue.

### DIFF
--- a/cloud/user.go
+++ b/cloud/user.go
@@ -257,6 +257,22 @@ func WithAccountId(accountId string) UserSearchF {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


elgranjero:
I need this feature too

PR #697, opened 2025-01-17
Fix bearerauth example

JamesSlocumIH:
#### What type of PR is this?
* documentation

The example for using a bearer auth token had two parameters flipped.
It should be (client, url) not (url, client)

N/A

### DIFF
--- a/onpremise/examples/bearerauth/main.go
+++ b/onpremise/examples/bearerauth/main.go
@@ -15,7 +15,7 @@ func main() {
 	tp := jira.BearerAuthTransport{
 		Token: "<persona-access-token>",
 	}
-	client, err := jira.NewClient(jiraURL, tp.Client())
+	client, err := jira.NewClient(tp.Client(), jiraURL)
 	if err != nil {
 		panic(err)
 	}


andygrunwald:
@JamesSlocumIH The current function definition looks like (URL, client): https://github.com/andygrunwald/go-jira/blob/48c5a07771b63cda984b1c0845a8199b7ad61d61/onpremise/jira.go#L87

Or am I missing something?

JamesSlocumIH:
Sorry I believe I got confused here:
- This is the [jira.NewClient(Client, Url)](https://github.com/andygrunwald/go-jira/blob/47d27a76e84da43f6e27e1cd0f930e6763dc79d7/jira.go#L72) I was calling.
  - My go.mod pulled in 1.16.0 which is the linked version above
  - Thus my local go code thinks it needs to be the other way around
- This example uses onpremise.NewClient(Url, Client)
  - But since I am using the NewClient version in 1.16.0, in doesn't recognize this new format

andygrunwald:
Yeah, it could be that we changed it since the last tag. Thanks for pointing it out.

PR #717, opened 2025-04-17
issue: switch to new search API

### LINKED ISSUE #715: Some APIs used will be shutdown on May 1st 2025

prarit:
#### What type of PR is this?

bug

Optionally add one or more of the following kinds if applicable:

api-change 

@dustin-decker reported that the current search APIs will be deprecated on May 1 2025 [1].

Switch to the new search API.

[1] https://developer.atlassian.com/changelog/#CHANGE-2046

Fixes #715 

This is untested.  I want to see if this passes CI.


ns-mglaske:
fwiw, SearchPages function @1166 is missing a `return nil` in it..

ns-mglaske:
Also, I think searchResult needs a `nextPageToken` field, in case you want to use Search and iterate it.

ns-mglaske:
Also, I think Response in jira.go needs the NextPageToken, along with updates to populatePageValues..

ns-mglaske:
Also.. It seems that a number of the fields that jira returns are no longer strings (eg, Description), but Jira's ADF format which really sucks :).  I wrote a adf_translator such that you can replace the IssueFields, eg, Description <string> with Description <ADFText>.  You could also take it a step further, and just do `type ADFText string` for a more direct replacement, but I wanted to know which fields broke in my upstream code.

```
package jira

import (
        "encoding/json"
        "fmt"
        "strings"
)

// ADFText represents an Atlassian Document Format field like "description",
// and extracts just the plain text content.
type ADFText struct {
        Text string
}

// UnmarshalJSON extracts plain text from the ADF document structure.
func (a *ADFText) UnmarshalJSON(data []byte) error {
        // Define a partial struct to match ADF layout
        var doc struct {
                Content []struct {
                        Content []struct {
                                Text string `json:"text"`
                        } `json:"content"`
                } `json:"content"`
        }

        // Unmarshal into ADF format
        if err := json.Unmarshal(data, &doc); err != nil {
                // Fallback: maybe it's just a plain string?
                var s string
                if err2 := json.Unmarshal(data, &s); err2 == nil {
                        a.Text = s
                        return nil
                }

                return fmt.Errorf("ADFText unmarshal failed: %w", err)
        }

        // Build plain text from nested content
        var sb strings.Builder
        for _, block := range doc.Content {
                for _, inline := range block.Content {
                        sb.WriteString(inline.Text)
                }
                sb.WriteRune('\n')
        }

        a.Text = strings.TrimSpace(sb.String())
        return nil
}

// MarshalJSON outputs the plain text as a JSON string (not as ADF)
func (a ADFText) MarshalJSON() ([]byte, error) {
        return json.Marshal(a.Text)
}

// ADFDocument represents the minimal ADF structure Jira accepts
type ADFDocument struct {
        Type    string       `json:"type"`    // always "doc"
        Version int          `json:"version"` // always 1
        Content []ADFContent `json:"content"`
}

func (a *ADFDocument) JSON() string {
        b, _ := json.Marshal(a)
        return string(b)
}

type ADFContent struct {
        Type    string        `json:"type"` // usually "paragraph"
        Content []ADFTextNode `json:"content"`
}

type ADFTextNode struct {
        Type string `json:"type"` // always "text"
        Text string `json:"text"`
}

// takes a plain string and converts it into an ADF JSON payload
func TextToADF(text string) *ADFDocument {
        // Split into lines → paragraphs
        paragraphs := strings.Split(text, "\n")
        var content []ADFContent

        for _, para := range paragraphs {
                para = strings.TrimSpace(para)
                if para == "" {
                        continue
                }

                content = append(content, ADFContent{
                        Type: "paragraph",
                        Content: []ADFTextNode{
                                {
                                        Type: "text",
                                        Text: para,
                                },
                        },
                })
        }

        return &ADFDocument{
                Type:    "doc",
                Version: 1,
                Content: content,
        }
}

func TextToADFJSON(text string) string {
        doc := TextToADF(text)
        b, _ := json.Marshal(doc)
        return string(b)
}
```

Jcparkyn:
@ns-mglaske I think you can avoid some of the effort of supporting ADF by sticking with the v2 version of the search endpoint.

The important part for the deprecation notice is switching from `/rest/api/2|3|latest/search` to `/rest/api/2|3|latest/search/jql`, but the `/jql` versions of the v2 endpoints (which don't use ADF) are still fully supported. IMO for this PR it would be better to prioritise that switch, and leave the migration to `v3` as a separate change since it's less urgent and harder to implement.

See https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#version

cloud/issue.go
```diff
@@ -1046,28 +1065,43 @@ func (s *IssueService) AddLink(ctx context.Context, issueLink *IssueLink) (*Resp
 // This double check effort is done for v2 - Remove this two lines if this is completed.
 func (s *IssueService) Search(ctx context.Context, jql string, options *SearchOptions) ([]Issue, *Response, error) {
 	u := url.URL{
-		Path: "rest/api/2/search",
+		Path: "rest/api/3/search/jql",
```
Jcparkyn, line 1068: As mentioned in PR discussion, it might be better to stick with `v2` for now to avoid the breaking changes from strings to ADF. V2 is still supported (not deprecated), provided that you use the `/jql` version which uses `nextPageToken`.

andygrunwald:
Hey all,
please see my latest comment in https://github.com/andygrunwald/go-jira/issues/715#issuecomment-3288768295 with an update on this one here.

PR #721, opened 2025-06-11
fixed issue where creating service desk request reponse was erroring out

aswinkumar-ar:
#### What type of PR is this?
bug

Creating a servicedesk request via the SDK fails to handle the response.
SDK assumes that value in RequestFieldValue is always string. But it is not the case.

Fixes #

This is the error I was getting while using the SDK.
```
2025/06/11 12:33:34 [DEBUG] POST https://support.solaris-testing.de/rest/servicedeskapi/request
{"level":"error","message":"Failed to create request: json: cannot unmarshal array into Go struct field RequestFieldValue.requestFieldValues.value of type string: file already closed"}
{"level":"error","message":"Failed to create request: json: cannot unmarshal array into Go struct field RequestFieldValue.requestFieldValues.value of type string: file already closed, response: "}

```
Below is a sample of requestFieldValues in the response from Jira API. As you can see value is not always string.

  ```
"requestFieldValues": [
    {
      "fieldId": "summary",
      "label": "Summary",
      "value": "Test issue finom"
    },
    {
      "fieldId": "description",
      "label": "Description",
      "value": "Creating an issue with the go-jira client"
    },
    {
      "fieldId": "attachment",
      "label": "Attachment",
      "value": [],
      "renderedValue": []
    },
    {
      "fieldId": "customfield_21700",
      "label": "",
      "value": []
    }
  ],
```

### DIFF
--- a/cloud/request.go
+++ b/cloud/request.go
@@ -26,7 +26,7 @@ type Request struct {
 type RequestFieldValue struct {
 	FieldID string `json:"fieldId,omitempty" structs:"fieldId,omitempty"`
 	Label   string `json:"label,omitempty" structs:"label,omitempty"`
-	Value   string `json:"value,omitempty" structs:"value,omitempty"`
+	Value   interface{} `json:"value,omitempty" structs:"value,omitempty"`
 }
 
 // RequestDate is the date format used in requests.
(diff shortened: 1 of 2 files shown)


aswinkumar-ar:
Is it possible to have this fixed in v1?
I thought it is not possible to fix it in v1. Hence closed.

PR #724, opened 2025-08-27
Fix Deprecation of JQL search and Evaluate expression endpoints

### LINKED ISSUE #715: Some APIs used will be shutdown on May 1st 2025

andygrunwald:
#### What type of PR is this?

* bug

The endpoint `GET /rest/api/2|3|latest/search` is being deprecated. This is the migration of the endpoint.

See https://developer.atlassian.com/changelog/#CHANGE-2046

Fixes #715

* Unit tests are not working as expected for now
* Testing by the community is needed
* Partially based on https://github.com/andygrunwald/go-jira/pull/717

### DIFF
--- a/cloud/board.go
+++ b/cloud/board.go
@@ -54,6 +54,9 @@ type BoardListOptions struct {
 	// Relevance meaning that the JQL filter defined in board contains a reference to a project.
 	ProjectKeyOrID string `url:"projectKeyOrId,omitempty"`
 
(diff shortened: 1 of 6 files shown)


moritzschmitz-oviva:
We did test this, and it works well, except for the issue mentioned in the comment above. Also see my original comment here: https://github.com/andygrunwald/go-jira/issues/715#issuecomment-3255139149.

PR #725, opened 2025-08-28
Add SearchV2JQL to fix deprecation of JQL search and evaluate expression…

### LINKED ISSUE #715: Some APIs used will be shutdown on May 1st 2025

conor-naranjo:
… endpoints

#### What type of PR is this?

* bug

The endpoint GET /rest/api/2|3|latest/search is being deprecated. This is the migration of the endpoint.

See https://developer.atlassian.com/changelog/#CHANGE-2046

Fixes #715

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -527,6 +527,63 @@ type SearchOptions struct {
 	ValidateQuery string `url:"validateQuery,omitempty"`
 }
 
+// SearchOptionsV2 specifies the parameters for the Jira Cloud-specific
+// paramaters to List methods that support pagination
+//
+// Docs: https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-search/#api-rest-api-2-search-jql-get
+type SearchOptionsV2 struct {
+	// NextPageToken: The token for a page to fetch that is not the first page.
+	// The first page has a nextPageToken of null.
+	// Use the nextPageToken to fetch the next page of issues.
+	// Note: The nextPageToken field is not included in the response for the last page,
+	// indicating there is no next page.
+	NextPageToken string `url:"nextPageToken,omitempty"`
+
+	// MaxResults: The maximum number of items to return per page.
+	// To manage page size, API may return fewer items per page where a large number of fields or properties are requested.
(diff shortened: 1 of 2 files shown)


erezrokah (commented):
Thanks for the PR @conor-naranjo added a couple of comments. Noticed them while testing the PR

erezrokah (commented):
Tested this approach a bit more, and it doesn't work as the v3 API returns a response with a different structure, so JSON unmarshalling fails. For example in v2 `fields.description` is a string and in v3 it's an object.

The Go struct expects a string.

A possible working solution is to use the v2 API, see https://github.com/cloudquery/go-jira/commit/9f996e429087a3d13e8012d57ef1d4d277aca31f

> Oddly enough Jira docs have it the other way around, v2 says it returns an object and v3 says it returns a string, but I confirmed v2 returns a string

image
image

mspraggs:
Hello :slightly_smiling_face: 

Thought Machine maintains a fork of this repo with some minor modifications to the v1 branch. Today we used the change in this PR as part of a separate PR [here](https://github.com/thought-machine/go-jira/pull/15) and found the changes worked well with our internal workloads. Just reporting back to let you know these changes work well "in the wild".

Many thanks

Matt

conor-naranjo:
@cxfcxf as I understand it, the new SearchOptionsV2 struct is to provide a cloud API-specific search options to maintain compatibility with on-prem because the v1 version of go-jira doesn't have a separate cloud vs on-prem client

conor-naranjo:
@cxfcxf that's what this PR is intended to address for the v1 version of the lib.
the v2 library (main branch -- not released) has separate JiraClient for cloud or on-prem with separate implementations

There's https://github.com/andygrunwald/go-jira/pull/724 to address fixing v2

andygrunwald:
Huge shoutout to @conor-naranjo for creating the PR, @erezrokah for the thorough review and @mspraggs for the "it works" confirmation.

I merged the PR and released [v1.16.1](https://github.com/andygrunwald/go-jira/releases/tag/v1.16.1). I will backport this PR into the current main version as well.

PR #739, opened 2025-09-26
correct return structure for GetList issuelinktype methods

svdaddio:
#### What type of PR is this?
bug

The `GetList` methods for the `IssueLinkTypeService` were expecting to see a return value that could be unmarshaled into `[]IssueLinkType{}` but the return is actually `struct{IssueLinkTypes []IssueLinkType}`

Fixes #
[Issue 740](https://github.com/andygrunwald/go-jira/issues/740)


antisedative:
Hi! Can you pay attention to this?
@andygrunwald

